Issue #1, opened 2016-02-24
Pass Flake8


ataffanel:
cflib is not passing the Flake8 tests yet, but we are not far!

The code should be fixed and flake8 should be added to ```tools/build/build``` to test in in CI.

ataffanel:
Fixed by commit 4642236

PR #2, opened 2016-02-25
Get closer to passing flake8

kentwills:
Not too sure about what you want to do with (pretty unfamiliar with the side-effects for the imports):

```
cflib/bootloader/__init__.py:37:1: F401 'Target' imported but unused
cflib/bootloader/__init__.py:40:1: F401 'Caller' imported but unused
cflib/crazyflie/toccache.py:36:1: F401 'LogTocElement' imported but unused
cflib/crazyflie/toccache.py:37:1: F401 'ParamTocElement' imported but unused
```

These changes should get you pretty close for #1


ataffanel:
Thanks! Yes I agree these last imports have potential side effects, I can look into them.

PR #4, opened 2016-02-26
DRY examples

### LINKED ISSUE #1: Pass Flake8

kentwills:
Tested what I could locally with a Crazyflie v1 using `python -m`, also closes #1.

### DIFF
--- a/cflib/bootloader/__init__.py
+++ b/cflib/bootloader/__init__.py
@@ -34,10 +34,10 @@ import time
 import zipfile
 
 from .boottypes import BootVersion
-from .boottypes import Target
+from .boottypes import Target  # noqa
 from .boottypes import TargetTypes
 from .cloader import Cloader
-from cflib.utils.callbacks import Caller
+from cflib.utils.callbacks import Caller  # noqa
 
 logger = logging.getLogger(__name__)
 
--- a/cflib/crazyflie/toccache.py
+++ b/cflib/crazyflie/toccache.py
@@ -33,8 +33,8 @@ import logging
 import os
 from glob import glob
 
-from .log import LogTocElement  # pylint: disable=W0611
-from .param import ParamTocElement  # pylint: disable=W0611
+from .log import LogTocElement  # noqa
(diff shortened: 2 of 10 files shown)


ataffanel:
I am sorry but I am not going to merge this one.

The examples are not dry by design: they are meant to be used as a starting point for other scripts and as such they should not "hide" details. Furthermore the test folder is not supposed to become an installed module so it cannot be imported.

I can push the last fixes to #1 and will link it to this PR.

Thanks

kentwills:
@ataffanel I would encourage you to think about this a little more.  There are a few things that are changed here:

# exit(main())
It is a good practice to wrap main in exit like: exit(main())

why? well if you ever want to test the main method for your python code, using sys.exit is just not going to work.  It is good to get in a habit of doing this.

# DRY and variable naming
Because the code was not DRY, there was a lot of copy and pasting going on, for instance:

https://github.com/bitcraze/crazyflie-lib-python/pull/4/files#diff-00ab5e9851997379814762dd3de60492L115

For the longest time I was trying to figure out what `le` was supposed to stand for.  I ended up looking at the other examples and determined it was pasted from examples/basiclog.py and stood for log_example.  I was further confused because the variable wasn't used after the reference.

Also, because the code wasn't DRY my eyes had to match on ~15 lines of code in each example instead of one function that I could immediately say, "oh yeah that is the connection string code for the crazyflie,"  I have seen this before and I understand it.  I don't believe placing it in another file, hides it from users.

If we were to use the DRY concept and used more expressive naming for variables I feel like the examples become much easier to parse.  Even if you don't agree with separating the code into a util file, we should probably make this a function at least. (so people can have the instant association from example to example)

"Furthermore the test folder is not supposed to become an installed module so it cannot be imported." -- I wasn't able to parse out what this comment was supposed to mean, could you elaborate?

PR #7, opened 2016-03-10
Add precommit to project

kentwills:
To see this in action:
1. ``virtualenv venv``
2. ``source venv/bin/activate``
3. ``pip install precommit``
4. ``pre-commit run --all-files``

Typically you will not type ``pre-commit run --all-files.``  Instead you will run ``pre-commit install`` and it will run automatically on any code in which you do ``git commit``.  More information [here](http://pre-commit.com/)

This is a great example where having a Makefile that automates creation of the virtualenv and ensures that hooks are installed is very nice to have.  I was going to add it to the Readme in the contribution section, but that is up to you.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,21 @@
+-   repo: git://github.com/pre-commit/pre-commit-hooks
+    sha: 616c1ebd1898c91de9a0548866a59cbd9f4547f6
+    hooks:
+    -   id: autopep8-wrapper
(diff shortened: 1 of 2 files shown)


krichardsson:
Nice!
I like the idea of running all these tools on commit. 
Even better would be to run them in the CI server. Maybe we should look at the possibility to add it to the docker builder?

Thanks!

kentwills:
@krichardsson yup, I will take a look at adding to CI and/or the docker builder

PR #8, opened 2016-03-11
Add precommit and tox for testing in travis

kentwills:
I believe this takes care of the three things that you were testing (bdist, tests, flake8)

In addition you have the added functionality:
``make venv`` will create a virtualenv with all requirements for the project
``make test`` will run tests locally
``make clean`` will clean up your environment

- tox will use unittest to test python 3.3 and 3.4 and travis will test both in parallel.
- pre-commit will check the commit for any linting errors
- travis no longer requires sudo.

I think the Dockerfile could still be a good idea if we want to run the environment on windows machines, but I don't think it is necessary for travis.

I am sure some changes might be controversial, based on what you had, so let me know and we can work through it.


krichardsson:
Thanks!

My view is that we should try to keep the build in the docker container, also on travis. I'll try to explain my thinking and why I think docker is the right tool.

1. It should be easy to reproduce the build environment locally. When a user builds (runs tests, analyses and so on) on his/hers local machine and the build passes, it should also pass on the build server. A common problem is that the versions of interpreters, frameworks and tools differs between development and build environment and lead to unpredictable results. A solution that keep them synchronized is preferable. 
1. Zero or little installation. There are tools in the python world (as you have shown) that handles versions and so on in a nice way, but we are also using other languages/tools/frameworks in other projects (ruby, php, C, java...). They all come with their own set of tools and solutions for versioning and handling environments. If we were only using one language, installing tools and setting up a environment might work out, but a common solution for all languages would be nicer. My hope is that by lowering the bar to get started, we can encourage more people to play with the code. At the same time I think it is important not to impose some specific tool chain on the user, that is if you want to install a full python environment on your machine that should work as well (and probably better than the general solution).
1. Decoupling from the build server. Travis is nice, but we might not use it forever and a generic solution will not tie us to travis.

My solution to this problem is using docker and the [toolbel](https://wiki.bitcraze.io/projects:dockerbuilderimage:index). It might or might not be the best solution? What are your thoughts?

That's the background, now to solutions :-)

Do you think it would be possible (or worth) implementing a similar functionality in a docker container? My python knowledge is fairly limited, but are there any tools around that can run tests on different environments that we could add the to builder image?

kentwills:
I'll try to address your points one by one:

## Docker
I love docker.  We use it at work for precisely the same reason you mentioned, reproducibility.  When the user runs tests they should run the same way on their machine as they do on a build machine.  The reason why I don't believe we need docker here is:
- Docker helps in isolating binary dependencies for reproducibility.  I don't see any here (let me know if I am missing anything)
- For python dependencies, we are gaining an isolated python environment (and are using an explicit version of python) through the use of a virtualenv without the need of docker because it works on all platforms
- I would also argue that `pip install tox` is much less of an installation requirement than ensuring that docker works locally on your machine
- Yes docker layers are cached, but rebuilding a docker every time you want to do a build (some layers will break the cache) is not fun and working inside of a docker precludes you from using all of the tools that you normally develop with on your system. 
- Docker needs to be installed on the system

I think using docker boils down to the fact that you are trying to make the build systems/ testing the same across all of the crazyflie projects to make it easy for people to get started.  I am looking at this as an isolated project in which to develop for and I believe that is the best way to look at the situation.  If I was contributing to a ruby project I would expect to use those tools, java, those tools.  In fact as a contributor, I would probably have a language of choice and would be comfortable/willing to learn how to use the common tools for that language.  We also try to make all the steps very easy to follow, ie to get started testing this package:
``pip install tox``
``tox``
If I am a python developer, I would be very comfortable with this environment.  If I was new to python I would be encouraged that this is what the community is using.  I would also not be wedged into using 'specific' python technologies.

One of my initial complaints with using the crazyflie client was that it required an elaborate install process or a vm to get up and running.  It took me quite a bit of time to get up and running.  With this package all I need is python installed on my system and I can immediately start developing for the crazyflie.

## Toolchain
Can you describe more what you mean here about installing a full python environment?
We test in a virtualenv because it will not mutate the global python state, but by no means is it a restriction here. 

## Travis
I have used both Travis and CircleCI.  Both provide support for Docker and both provide python and tox support.  CI will support toolchains for whatever language you choose.  This is another reason that I vote for using the toolchain for the language.  We are, by no means, tying ourselves to travis with the travis file I have created.

## Answers to your questions
Do you think it would be possible (or worth) implementing a similar functionality in a docker container? 
Possible, but I still don't see the benefit that we gain here based on comments above.

My python knowledge is fairly limited, but are there any tools around that can run tests on different environments that we could add the to builder image?
Not quite sure what we are talking about here.  Are we talking about the Docker Image? What environments are we talking about, linux environments, language environments?  If we are talking about python versions, tox covers that.  If we are talking about linux versions, CircleCI does a good job at running your tests on different linux distributions in a parallelizeable manner.  If we are talking about osx, windows and linux, yes I think we can do that as well, but the code will be different per CI platform that we would want to go with.

## Final Message
I can help you do whatever you want to do for this project.  I just want you to understand that some of the choices that are being made may, in fact, increase the complexity of the project because of the global perspective that you are thinking about things in.

krichardsson:
I don't really think there are any conflicting interests here. What I suggest hopefully supports different ways of working, ranging from installing a full python environment with your favourite IDE fully integrated, to a very simplistic approach requiring only docker and a text editor. I definitely don't want to impose any specific environment or set of tools on the developer, rather the other way around.  

# Docker
First I will explain the way we use docker in this use case (as it is a versatile tool and can be used in manny ways) to avoid misunderstandings. 
What we do here is to use docker to encapsulate tools. Further more the containers have a very short life cycle of only one operation, for instance running the tests. 
What happens is 

1. A builder container is started with the project files mounted in the container
1. The task is executed in the container, operating on the mounted files
1. The container is terminated and removed

The container is based on a builder image from Bitcraze (bitcraze/builder) so there is no image to build for the user, only pulling it the first time it is used. Starting the container is instantaneous and not noticeable when used on the command line. There is a slight penalty in speed on OSX since the mapping of the filesystem is not that good.
The toolbelt is mainly a tool to make it easy to run these short lived containers.

# Travis
To me travis adds a number of useful things

1. Automatic build on updates to the repository
1. The ability to run tests on multiple versions and environments
1. Publishing build results and status
1. The definition of a successful build

1., 3. and 4. are generic CI/CD server functionality, while 2. is travis specific.

# Questions
I was probably not very clear in my questions, sorry!

What I'm interested in is if .2 of the travis functionality could be implemented in some other way? Maybe we could have a docker image with the tools and versions needed (venv?, tox?, various versions of python, ?) and a script that runs the builder container with the appropriate settings to test the variations we are interested in. This would also enable us to run all tests, for all versions locally.

# Conclusions
I hope I have managed to explain what I'm looking for. I agree that the complexity of the project increases slightly, but only in the automated build part. We will still support a pythonic way of working to cater for the experienced developers that expects a "normal" python environment, and will also enable less python oriented people to contribute.

kentwills:
To address point 2:

# Multiple Python Environments

- tox automates and standardizes testing in python using virtual environments
- virtualenv creates isolated python environments

Your choice is to depend on the above tools, create your own scripts, or rely on a particular CI tool.  I think there is one way to go forward without relying on CI infra.

## The above tools
Becomes an implementation detail that can be used in any CI environment.  Those that have used tox before will be able to immediately understand how things work, those that haven't just have to type the command tox.

## Your own tools
Becomes an implementation detail that can be used in any CI environment.  The scripts will be unfamiliar to others and will require people to read the code or read the docs.  It is unclear how you automate testing with multiple python environments with current scripts without relying on CI infrastructure.

## CI infrastructure
Give in, tie yourself to a CI platform:
- [TravisCI multiple envs](https://docs.travis-ci.com/user/languages/python)
- [CircleCI multiple envs](https://circleci.com/docs/language-python)

# Operating System Environments
If you are also concerned with operating system environments, then this is something to consider:

## using docker
You are essentially only testing in linux, but have support for all platforms because of docker compatibility on all platforms.  You also have a ground truth environment for reproducing errors (this is really less important for this package since it has no external binary dependencies).  You don't have to set environment variables in CI platforms for the OS;  Then again you are not proving that you are providing native support for anything but linux.  If someone wants to run their tests they can run the python script, but there is a hidden docker dependency here and they have to make sure that it is properly installed on their machine.

*If you only care about providing native support for linux, then I could see you going this direction.*

## not using docker
You can prove that you have native support on OSX and linux with Travis or Circle (requires two configuration lines, hardly ties you to a CI framework) and Windows support with Appveyor.  Furthermore, an individual contributor could run tox natively with confidence, because you have tested for those three scenarios.  They also don't have to deal with the hidden docker dependency. (tox dependency `pip install tox` = less work than [ensuring docker is installed](https://docs.docker.com/engine/installation/))

*If you care about natively supporting OSX, Linux, and Windows, I would suggest this way forward.*

## My Questions

- What would you like to see from me so that I can create an acceptable PR that will incorporate pre-commit and testing?
- Do you want me to run the tox command in docker?
- Do you want me to run tox outside of docker and also run the docker test?
- Do you want me to not use tox at all and stick to creating a script that can be run in the bin directory?

krichardsson:
Again, I don't really think there is an either or, why not both? 

I like your point that travis can test on multiple OSes, even though its not part of the config yet. That makes it worth using the "native" travis config for me.

I still think it is valuable to be able to run static code analyses, tests and so on in a docker container for the users that like that solution. My thinking is that it would run the checks in one environment only, the one set up in the docker image. What is needed for that is
1. A bash or python script in the tools/build dir that calls the appropriate tools to run tests and so on
1. The tools that are used by the above script must be added in the docker builder image, if missing (defined in https://github.com/bitcraze/docker-builder)
1. If possible, let travis run the tests in the docker container as well as a part of the build job. This would ensure that the docker images is kept in sync over time. I'm not sure if this can be done in the same job, but I think it is possible.

kentwills:
Ok will try to find a solution that supports both.  Will let you know when it is ready to be looked at again.

kentwills:
Seems like travis has some more work to do in order to support python, https://github.com/travis-ci/travis-ci/issues/4729

kentwills:
also https://travis-ci.org/kentwills/crazyflie-lib-python/jobs/117624042

kentwills:
Ok I think this is the best way that I can support the docker build, the osx build, and the tox 27,33,34 build.  This last run on travis should be sufficient.  Let me know what you think.

krichardsson:
Merging 
Thanks  for your patience and hard work!

PR #13, opened 2016-04-26
Add support to enable/disable acknowledgements.

whoenig:
This also ensures that the Crazyradio is reset properly (initializes to acknowledgement enabled by default).


ataffanel:
Merged, thanks!

PR #14, opened 2016-05-20
Fix warmboot operation for python3.

whoenig:
Tested using cfloader -w radio://0/100/2M/E7E7E7E701 flash cf2.bin stm32-fw


krichardsson:
Merged
Thanks!

PR #16, opened 2016-09-20
don't list debug drivers in examples

whoenig:


ataffanel:
Thanks! the code now match the comments :-)

PR #21, opened 2016-11-11
Add example to scan for good radio channels

AiRuiChen:


ataffanel:
Thanks a lot, the pull request looks good!

Though I have problems getting the script to work, do I need any specific version of the Crazyflie nrf firmware or of the Crazyradio?. The scripts outputs something like that:

```
$ python3 radio-test.py -channel 70 -rate 2
/home/arnaud/.local/lib/python3.5/site-packages/numpy/core/_methods.py:59: RuntimeWarning: Mean of empty slice.
  warnings.warn("Mean of empty slice.", RuntimeWarning)
/home/arnaud/.local/lib/python3.5/site-packages/numpy/core/_methods.py:70: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
/home/arnaud/.local/lib/python3.5/site-packages/numpy/core/_methods.py:82: RuntimeWarning: Degrees of freedom <= 0 for slice
  warnings.warn("Degrees of freedom <= 0 for slice", RuntimeWarning)
/home/arnaud/.local/lib/python3.5/site-packages/numpy/core/_methods.py:94: RuntimeWarning: invalid value encountered in true_divide
  arrmean, rcount, out=arrmean, casting='unsafe', subok=False)
/home/arnaud/.local/lib/python3.5/site-packages/numpy/core/_methods.py:116: RuntimeWarning: invalid value encountered in double_scalars
  ret = ret.dtype.type(ret / rcount)
Channel 0 ack_rate: 0.0 rssi average: nan rssi std: nan
Channel 1 ack_rate: 0.0 rssi average: nan rssi std: nan
Channel 2 ack_rate: 0.97 rssi average: 37.0 rssi std: 0.0
Channel 3 ack_rate: 0.63 rssi average: 37.0 rssi std: 0.0
Channel 4 ack_rate: 0.37 rssi average: 37.0 rssi std: 0.0
Channel 5 ack_rate: 0.21 rssi average: 37.0 rssi std: 0.0
Channel 6 ack_rate: 0.0 rssi average: nan rssi std: nan
Channel 7 ack_rate: 0.0 rssi average: nan rssi std: nan
Channel 8 ack_rate: 0.0 rssi average: nan rssi std: nan
Channel 9 ack_rate: 0.0 rssi average: nan rssi std: nan
```

So, it looks like it is working for the first channels but then it just leaves the Crazyflie behind.

whoenig:
Is your Crazyflie far away? You might need to increase the loop count in lines 59 and 89 from 20 to 50. The warning comes from trying to compute the mean and standard deviation of an empty array. I guess we could add a guard for that case (even though in this case the output of nan is correct).

There shouldn't be a special firmware required. We tested with stock firmware and the crazyswarm firmwares.

ataffanel:
Yes I was close but I think the problem was coming from the fact that I have bluetooth enabled: the BLE stack is taking over the radio for some time and 20 was not long enough to pass this time. With 50 it works but the bluetooth stack induces pretty big "dips" in the measurement at regular intervals. Anyway it works so I will just set the default retry to 50 and add a comment about the bluetooth stack.

I added a ticket about the stack: https://github.com/bitcraze/crazyflie2-nrf-firmware/issues/17. When we can disable it at runtime we can add a 'magic packet' to disable bluetooth from the air.

Thanks for the PR!

Issue #75, opened 2018-04-02
How to set_arc for crazyradio using crazyflie class?


barzanisar:
Hi,
I am using [cfbridge](https://github.com/dennisss/cfbridge/blob/master/scripts/cfbridge.py) to connect to crazyflie 2.0 from QGC via crazyradio PA. 

Cfbridge creates a crazyflie object that connects to crazyflie using [radiodriver.py](https://github.com/bitcraze/crazyflie-lib-python/blob/master/cflib/crtp/radiodriver.py)

How can I configure the radio's number of retries i.e. in [here](https://github.com/bitcraze/crazyflie-lib-python/blob/master/cflib/crtp/radiodriver.py#L196) from crazyflie class? 

I want to be able set the number of retries in my cfbridge rather than hard coding it in radiodriver.py

ataffanel:
There is currently no way to achieve that without modifying the radiodriver code.

I think this case if very similar to the ticket #58 and should be solved the same way so that you could set the number of radio retry as a global option before connecting the Crazyflie.

ataffanel:
The PR is merged so I close this ticket. Thanks!

PR #76, opened 2018-04-04
set radio retry as global option

barzanisar:
This is done to solve issue https://github.com/bitcraze/crazyflie-lib-python/issues/75


ataffanel:
Thanks! Merged.

Issue #89, opened 2018-08-07
Implement new log/param command to handle more than 255 log/param variables


ataffanel:
bitcraze/crazyflie-firmware#337 implements new ctrp commands to access more than 255 log/param variables. Support for it should be implemented in the lib.

PR #93, opened 2018-08-24
Continue to support V1 protocols.

simonjwright:
A couple of cases were missed in updates for #89, such that comms fail if the crazyflie is still running protocol v1 (not sure if that’s the right name? anyway, the previous protocol).
 
  * cflib/crazyflie/param.py (Param._param_updated): if not using
      protocol v2, only strip one byte before unpacking the data.
  * cflib/crazyflie/toc.py (TocFetcher._new_packet_cb): likewise, for
      GET_TOC_ELEMENT.


ataffanel:
Thanks for the PR! this was indeed missed-out. I merge it.

PR #94, opened 2018-08-27
docs: addition to Testing section about python3.6 on linux

Ayshine:
I tried to install tox with `pip install tox` my testings failed but
`sudo apt-get install tox` worked and using the command `TOXENV=py36 tox`
my text passed. Thus, I am **proposing** add those line below to the
 README.md file

* on Linux for python3.6 if the test fails after installing tox with
 `pip install tox`, installing with  `sudo apt-get install tox` result
 a successful test run

* Test package in python3.6 `TOXENV=py36 tox`

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,11 @@ work needed to maintain your python environmet.
 
 ### Native python on Linux, OSX, Windows
 * [Tox](http://tox.readthedocs.org/en/latest/) is used for native testing: `pip install tox`
+* on Linux for python3.6 if test fails after installing tox with `pip install tox`, installing with  `sudo apt-get install tox`result a successful test run
+
 * Test package in python2.7 `TOXENV=py27 tox`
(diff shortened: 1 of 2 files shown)


ataffanel:
Thanks for the readme change, it looks good to me!
Though, I do not understand and I do not think we want to merge the other two changes in this commit. I assume the src/ folder is created by tox when running?

Ayshine:
Thanks, @ataffanel! My intention was the only add two lines to the README file. You are correct, the other two changes were not intended. I am new to crazyflie platform so I couldn't find the other two changes before sending a pull request. Can we exclude them ? or should I try to submit a new pull request only for the README file?

Ayshine:
I'm closing this pull request and will open just for README file :)

ataffanel:
Yes that would have been my suggestion, thanks!

PR #97, opened 2018-09-01
docs: changes to documentation

Ayshine:
On an Ubuntu system following two lines worked for successful installation of tox and test passed. As a result, I am proposing a change to README.md documentation by adding two lines below. 
* If test fails after installing tox with `pip install tox`, installing with  `sudo apt-get install tox`result a successful test run

* Test package in python3.6 `TOXENV=py36 tox`


ataffanel:
Thanks for the PR! I am merging it.

PR #125, opened 2019-08-31
Updated example for multiranger to use pyQt5

arjo129:
PyQt4 and Qt v4 are no longer supported and no new releases will be made. Just updated the example to use PyQt5.


krichardsson:
Thanks!

PR #129, opened 2019-11-15
Append the device only if the manufacturer name is 'Bitcraze AB'

ashwinsushil:
When devices with the same VendorID and ProductID as the
roadrunner/crazyflie are connected at the same time, the cflib fails to connect to
the device. A quick fix for this is done by checking if the
manufacturer name is 'Bitcraze AB'

### DIFF
--- a/cflib/drivers/cfusb.py
+++ b/cflib/drivers/cfusb.py
@@ -65,7 +65,8 @@ def _find_devices():
     if pyusb1:
         for d in usb.core.find(idVendor=USB_VID, idProduct=USB_PID, find_all=1,
                                backend=pyusb_backend):
-            ret.append(d)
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks for the fix! I am merging it.

ashwinsushil:
Heads up guys, if the manufacturer name is changed in the future, this fix wouldn't work. In fact, it will not even connect to any devices. I think we should think about a robust way to find the devices.

ataffanel:
Yes agreed, the real solution would be to have out own VID/PID but getting a VID is currently not possible. Maybe ST can help with that ....
In the meantime fixing the manufacturer string is the best we can do to make sure we do not open other devices using this vid/pid.

PR #133, opened 2019-12-17
Increase write flash request timeout to 2.5s, fixes #98

### LINKED ISSUE #98: Bootloader is sometime not flashing reliably
There has been more occurrence of the Bootloader flashing a corrupt firmware which prevents the Crazyflie from starting and requires to flash the firmware again. The bootloader protocol is designed to be safe so there is a strong assumption that something has been broken in the python bootloader client implementation.

ntamas:
As discussed in #98, this PR raises the timeout for "write flash" requests to 2.5 seconds unconditionally to cater for the increased time needed to erase a page of 128K on an STM32F405.

Note that the timeout is applied unconditionally, even if we are flashing to the nRF51. In my opinion this is not a big deal because the flash packets go through just fine almost all the time, so there is almost never any need for re-sending them. However, re-sending them too early when the packet in fact went through can break the flashing process, as indicated by #98.

### DIFF
--- a/cflib/bootloader/cloader.py
+++ b/cflib/bootloader/cloader.py
@@ -381,7 +381,14 @@ class Cloader:
             pk.data = struct.pack('<BBHHH', addr, 0x18, page_buffer,
                                   target_page, page_count)
             self.link.send_packet(pk)
-            pk = self.link.receive_packet(1)
+
+            # Timeout for writing to flash is raised from 1s (used elsewhere
+            # in this module) to 2.5s because it may take more than a second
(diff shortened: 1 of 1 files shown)


ataffanel:
Thanks a lot, this is going to help a lot of people and save a lot of time!
I merged the PR, lets keep the issue open a little while to make sure the problem is indeed gone.

ntamas:
I'm fine with keeping the issue open but you just closed it ;)

ataffanel:
Thanks for the heads-up, it was automatically closed by github because of this PR title :-).

PR #141, opened 2020-02-26
prrt functionality

s9meboeh:
Enable CRTP communication via PRRT.


knmcguire:
This is accommodating the pull request #140, so this needs to be merged together with this one. Using PRRT is maybe for now not intergrated into the library completely, as it mostly is used for the raspberry pi intergration, but it might come in handy for the future.

PR #157, opened 2020-07-01
Fix documentaion and some code typos

joshmeranda:
Fix typos in project documentation, docstrings, and some code segments. There are still a few remaining misspelled methods and argument names; however, changing them _may_ break backwards compatibility for some dependents, but they're readability is not largely impacted.

While the scope of this PR is quite large, the impact to anything other than readability should be nonexistent so I have not created an associated issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ For information and installation of the
 * Check to see if you pass style guidelines: `tb verify`
 
 Note: Docker and the toolbelt is an optional way of running tests and reduces the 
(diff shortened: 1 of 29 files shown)


knmcguire:
Hi! Sorry for the late reply, but indeed this is very helpful. I had a quick check and it all seems fine to me. Probably the define RETRYCOUNT_BEFORE_DISCONNECT should be removed completely, but ill make a seperate push for that.

Thanks a lot!

PR #158, opened 2020-07-15
Add memory handler for the Led Timings driver #597 + example

dolfje:
As we introduce the functionality of a led timings driver, we update the library to support this new led method.
And also add an example.


knmcguire:
The build has failed so that would need to be fixed. You can check it in the details of travis CI or if you have the toolbelt installed you can run test locally on your machine. https://github.com/bitcraze/crazyflie-lib-python#testing

Let us know if you need help with that.

cflib/crazyflie/mem.py
```diff
@@ -194,6 +197,68 @@ def disconnect(self):
         self._write_finished_cb = None
 
 
+class LEDMemoryDriverMemory(MemoryElement):
```
knmcguire, line 200: What is the difference between this one and the class LEDMemoryDriverMemory ? The explanation says exactly the same function. Why is the already existing one not sufficient for this or can be adjusted to do what you want it to do?

dolfje: LEDDriverMemory is handling the memory to show a fixed static pattern across the leds of the deck,
while LEDMemoryDriverMemory is handling the memory to show a changing pattern across the leds of the deck. So possibility to upload a show of multiple led patterns and from the moment you enable the lights, it will go through that show pattern by patterns with the ability to specify if the next pattern should fade in and what leds that should be enabled.

knmcguire: Ah yeah oke I understand. But currently the explanation of this class is a direct copy:
    _"""Memory interface for using the LED-ring mapped memory for setting RGB
       values for all the LEDs in the ring"""_
This should be adjusted to explain the difference.

knmcguire: In general, it might be better to have a different name for it. LEDSequenceDriverMemory perhaps? Or just merge it with the LEDdrivermemory. We kind of want to avoid duplicates if you can understand

dolfje: Is something like LEDTimingDriverMemory good? As the memory layouts are totally different, I think it is best not to merg with LEDDriverMemory

knmcguire: Yeah something like that :)

knmcguire:
Alright looks good! I will use this change to check out the pull request in the firmware and to try out the example. No time today anymore but I will try tomorrow.

knmcguire:
I'm currently trying both pull request on the led timing out at once. With running the python example, I get the following error message:

```
$ python3 examples/basicLedTiming.py 

Connecting to radio://0/80/2M/E7E7E7E7E7
Connected to radio://0/80/2M/E7E7E7E7E7
Traceback (most recent call last):
  File "examples/basicLedTiming.py", line 56, in <module>
    mem[0].add(25, {'r': 100, 'g': 0, 'b': 0})
IndexError: tuple index out of range
```

dolfje:
That looks like your crazyflie doesn't has the correct firmware. As the library can't find the memory module added in the firmware.

knmcguire:
hmm maybe not, but the firmware I flashed did had your commits in there I checked with git log.
'''
commit 787cc9bd1d1c5f48ab853a2654cc423622a901e2
Author: Nikos Verschore <nikosverschore@gmail.com>
Date:   Wed Jul 15 23:35:30 2020 +0200

    Test and fix the implementation of Led Timings
'''
also handleLedMemRead is in mem.c.

There might be something else going on

knmcguire:
is there maybe something in config.mk that needs to be added?

dolfje:
No changes to config.mk are needed. In the firmware itself it adds the new driver: https://github.com/bitcraze/crazyflie-firmware/pull/598/files#diff-a175f25f456b304028ac704ac00ee2b8R245 If both the crazyflie has the pull request firmware and you have installed the pull request library in your pc, then it should work. I just checked it with both branches without any configuration changes. Maybe you didn't reinstall the library on your pc?

But in general, have no clue, why it don't work with you :(

knmcguire:
so I cloned the repos on a temp folder to make sure its super fresh. The cflib is installed from that folder which i checked with pip3 lib:
`cflib                         0.1.11              /home/kimberly/projects/temp/crazyflie-lib-python`
 lib is in feature/597 from your repo. 

I did the same with the firmware which I checked with the cfclient in help>debug:
```
Crazyflie
Connected: usb://0
Firmware: 787cc9bd1d1c (CLEAN)
```
This is the last commit of your pull request.

Unfortunately... the same problem: 
```
$ python3 basicLedTiming.py 

Connecting to radio://0/80/2M/E7E7E7E7E7
Connected to radio://0/80/2M/E7E7E7E7E7
Traceback (most recent call last):
  File "basicLedTiming.py", line 56, in <module>
    mem[0].add(25, {'r': 100, 'g': 0, 'b': 0})
IndexError: tuple index out of range
```

Adding the following helps with the error, but it doesn't activate the leds:
`        if len(mem) > 0:`

knmcguire:
Next week one of the other developers are back so I will ask him to try it out on his computer. I still think there is something missing but at least we can be sure.

krichardsson:
Merged. Thanks!

PR #188, opened 2021-01-27
added uSD memory type support.

noamco3:
added support for uSD memory type as well as an example.

the transfer works, but the data transfers very slowly. I don't know if there's a way to speed it up, but the transfer itself works.

also added example code that turns off the uSD logging parameter.


tobbeanton:
Hi!
This is very nice functionality to have, nice work! 
I checked the pull-request and overall it looks good but I would like you to update the comments were many are left from "copy pasting" a class. Also please update the problems the CI finds and there are two .bin files that I guess shouldn't be in the pull request?

I think I know why it is slow to read out data and I is most likely due to that the FW opens and closes the log file at every read. Maybe this could be optimized somehow e.g. by transferring the whole file at once which might trigger many changes to achieve though...

knmcguire:
Hi! Since the build is failing and @tobbeanton 's request are not met yet, this PR is not ready to merge. As it is probably outdated, I'll close this PR, but please reopen or start a new one if you would like to contribute it still.

Issue #196, opened 2021-02-19
Add support for crazyflie-link-cpp


whoenig:
We developed a new link library in C++ for lower latency, higher bandwidth, and lower CPU utilization in https://github.com/bitcraze/crazyflie-link-cpp. This library already comes with Python bindings to be used with Python 3.x.

We first need to add support in crazyflie-lib-python, that can be enabled on demand using environment variable `USE_CFLINK=cpp`. Once we can confirm that it works well on all target platforms, we can switch to this backend by default.

jonasdn:
What actions are left before this can be closed?

whoenig:
This hasn't been tested thoroughly on all platforms yet. Ultimately, the goal was to enable this backend by default and possibly remove the older pure Python implementation in order to minimize combinations that we have to test.

jonasdn:
How do we reach the level of testing we need? Can we put-out a call for testing? Do we have go-to persons? Or should we just do it ourselves? Are there areas you are more worried about than others?

whoenig:
We already know that there are crashes on shutdown on Windows in the lib, so this needs to be fixed first. Otherwise the normal release-cycle testing will suffice (including our automated system tests, which we only ran from a Linux host so far).

ntamas:
Did you experience any crashes on shutdown on other platforms? In particular, I use an async-await-based alternative Crazyflie library and I was trying to make it work with `crazyflie-link-cpp` as it would simplify things a great deal when crossing from coroutine-world to thread-world and back (as the threading part would nicely be hidden in the C++ layer). I work on macOS and I get crashes like this upon exiting the Python interpreter:

```
Assertion failed: (pthread_mutex_destroy(mutex) == 0), function usbi_mutex_destroy, file ./os/threads_posix.h, line 58.
```

It comes from the `libusb_exit()` call and somehow it seems not to be consistent; I cannot reproduce it with simple scripts, only with a more complex test case that imports lots of stuff from my library in order to do a throughput test. I was just wondering whether this assertion failure is the same as the one you see on Windows.

The full stack trace of the crashed thread is as follows:

```
Thread 0:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff68a5a33a __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff68b16e60 pthread_kill + 430
2   libsystem_c.dylib             	0x00007fff689e1808 abort + 120
[...]

PR #197, opened 2021-02-19
Fix argument name of RadioDriver.receive_packet

whoenig:
The original definition is that function is to have an argument called
wait, not time, see https://github.com/bitcraze/crazyflie-lib-python/blob/master/cflib/crtp/crtpdriver.py#L59

Part of issue #196.


ataffanel:
Looks good, thanks!

PR #198, opened 2021-02-19
Bootloader: more explicit about address and safelink requirement.

whoenig:
Notes:
* The current radiodriver is not enforcing safelink, but
rather lazily initializing it. This change does not change the behavior
and the [noSafelink] hint in the URI is simply ignored.
* The current scan_selected function will only consider channel and
datarate. While more can be passed in (e.g., address), this is ignored.

Part of issue #196.


ataffanel (approved):
Looks good!

PR #204, opened 2021-03-04
radiodriver: Make sure we pad the address if less than 10 chars

jonasdn:
fixes bitcraze/crazyflie-clients-python#475


ataffanel:
Looks good, merging, thanks!

ataffanel (commented):
Looks good!

PR #212, opened 2021-03-16
Remove the debug driver

jonasdn:
The debug driver is a software (Python) endpoint for the CRTP protocol
that can be used to develop against when there is no accessable
hardware.

It requires maintenance in order to stay up-to-date with the CRTP
protocol and the Crazyflie and the benefit we receive from it does not
outweigh the amount work needed to keep it relevant.


whoenig (approved):
Nice work! User code will keep working, but a warning is printed. Examples are updated to serve as a good template.

PR #213, opened 2021-03-18
Make system_type in LighthouseConfigWriter.write_and_store_config() optional

krichardsson:
LighthouseConfigWriter.write_and_store_config() is used to write geo, config and system type data to a Crazyflie. All arguments should be optional to enable a client to chose what to write. Currently the system_type is set to LH V2 by default, but should be changed to None. The behaviour should also change so that no system type is set in the CF if system_type == None


knmcguire:
Looks good! After the check I'll merge it

PR #215, opened 2021-03-22
Fix regression where power_switch does not work with Python backend

whoenig:
This was introduced in 730aea39b321ddb78dfbfd6853563c1e7e134167, where
the Python backend accepted the URI options. For the power switch, we
still need an additional check to enable the old code path using the
shared radio.


ataffanel:
This is a good fix for now, in the future lets implement the link options in the python link or retire the python link. I created issue #218 to track this. In the meantime I am merging this, thank you!

PR #216, opened 2021-03-23
Add emergency stop and emergency stop watchdog support.

whoenig:
Emergency stop packets were added to the firmware May 2017
(commit 1e9c08c0ffc942dd0999f8b8dbbecfdfec2a84c7, firmware issue 226),
but they are not supported in the lib, yet. This change adds
send_emergency_stop and send_emergency_stop_watchdog to the
Localization class.


ataffanel (approved):
Good low level interface.

PR #220, opened 2021-03-23
Qualisys example: add support to only send position estimate.

whoenig:


krichardsson:
Looking good. I like that the full pose is an extra option to make it more stable by default

PR #221, opened 2021-03-24
System tests: improve documentation and reporting

whoenig:
* Add Readme that explains how to run the system tests.
* Update firmware from release zip, rather than separate bin files.
* Enable verbose output for sys-test script.


krichardsson:
Looks good to me. Nice improvement :-)

PR #222, opened 2021-03-24
Harmonize URI and address handling for examples

jonasdn:


ataffanel (commented):
I like the uri_helper uri_from_env and address_from_end but I do not like the helper being dependent of the radio driver.

I do not see a big need for separating the uri parsing in a separate file since it is very tightly coupled to the crtp radio driver.

jonasdn:
> I like the uri_helper uri_from_env and address_from_end but I do not like the helper being dependent of the radio driver.
> 
> I do not see a big need for separating the uri parsing in a separate file since it is very tightly coupled to the crtp radio driver.

Agreed! I forced-pushed an update that does not mess with the radio driver and instead just splits out the address part.

ataffanel (approved):
Looks good!

krichardsson:
Nice cleanup, I have been wanting to do something like this for a long time! 

There seems to be some imports that need to be reordered though, the build fails. Otherwise I think it is all good.


Another related thought is to maybe remove all the `cflib.crtp.scan_interfaces()` calls as well? They are a bit magic to me and adds complexity that makes it harder to understand what the example actually is about. This is probably another PR though, later on.

jonasdn:
> Nice cleanup, I have been wanting to do something like this for a long time!
> 
> There seems to be some imports that need to be reordered though, the build fails. Otherwise I think it is all good.
> 
> Another related thought is to maybe remove all the `cflib.crtp.scan_interfaces()` calls as well? They are a bit magic to me and adds complexity that makes it harder to understand what the example actually is about. This is probably another PR though, later on.

I am all for removing the scan_interfaces calls. And yes I noticed the CI fail... but I cannot reproduce it `tools/build/verify` passes for me :)

jonasdn:
Is `CFLIB_URI` the best name for the default environment variable?

Other options from the top of my head:

 * `CRAZYFLIE_URI`
 * `CRTP_URI`
 * `CRAZY_URI`
 * `CF_URI`
 * `CRAZYRADIO_URI`
 * `CRAZYRADIO`

krichardsson:
Wolfgang had a similar problem the other day, I have to look into what it might be

jonasdn:
> Wolfgang had a similar problem with the toolbelt the other day, I have to look into what it might be

Could it be related to this being a draft PR?

whoenig:
> Wolfgang had a similar problem with the toolbelt the other day, I have to look into what it might be

Mine was actually not a problem. The failure happened because master itself was broken. Once master got fixed, my PR turned green also (I think I had to trigger a manual re-run).

krichardsson:
The toolbelt problem seems to be related to the python tooling and not the toolbelt and docker. It seems as the checks are not run on all files (some sort of optimization I guess) and locally it skips test/crazyflie/test_syncLogger.py, test/crazyflie/test_syncCrazyflie.py and examples/tuning/PID_controller_tuner.py while they are checked on the server. If you change something in those files locally (like adding a line) and run `tb verify` it will check them and reorder the imports.

I have not idea of how to solve this problem in a good way

PR #224, opened 2021-03-29
Crazyradio server link driver

ataffanel:
This PR implements an experimental link driver for the experimental [Crazyradio Server](https://github.com/ataffanel/crazyradio-server).

It works for all use case, running the client, the examples and using the bootloader. But is not thoughrougly tested and has a couple of limitations:
 - The bootloader do no fail gracefully if no bootloader is detected
 - There is no way to setup the address of the Crazyradio server, it is hard-coded to "tcp://localhost:7777" for now.

This should be good enough for running examples and ```make cload``` in Windows in WSL while running the Crazyradio server as a windows application.

To use this link one should:
 - Run the server
 - Either:
    - Run scripts by prepending them with "USE_CFLINK=zmq", for example ```USE_CFLINK=zmq make cload```
    - Add ```export USE_CFLINK=zmq``` to the file ```~/.bashrc``` or ```.profile``` and then all scrips and "make cload" will use the Crazyradio server automatically.

### DIFF
--- a/cflib/crazyflie/mem/deck_memory.py
+++ b/cflib/crazyflie/mem/deck_memory.py
@@ -117,8 +117,12 @@ class DeckMemory:
     def _parse(self, data):
         self._bit_field = struct.unpack('<B', data[0:1])[0]
         if self.is_valid:
-            self.required_hash, self.required_length, self._base_address, _name = struct.unpack('<LLL19s', data[1:])
-            self.name = _name.split(b'\x00')[0].decode()
+            try:
+                self.required_hash, self.required_length, self._base_address, _name = struct.unpack('<LLL19s', data[1:])
+                self.name = _name.split(b'\x00')[0].decode()
+            except Exception as e:
+                logger.warning(f"Error while decoding deck mem ({e}), skipping!")
+                self._bit_field = 0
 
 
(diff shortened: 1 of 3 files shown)


knmcguire:
Since I'll be experimenting with ROS in WSL, it would be nice to have this integrated :)

What is the current status of this @ataffanel ?

whoenig:
I don't think this will work with ROS? Crazyflie_cpp uses libusb directly. Perhaps if we find a way to get this approach to work with the native-link we could get it to work with ROS(1). Otherwise, I'd say: skip WSL and use ROS2 natively on Windows:-)

knmcguire:
true.. but since ignition is not fully ported to windows yet I'm still a bit stuck on WSL2 with an xserver for that, or the native ubuntu install.

Interesting thing is that microsoft released the [ following WSL usb sharing tool: ](https://devblogs.microsoft.com/commandline/connecting-usb-devices-to-wsl/ 
). It works also for windows10 but crazyradio crashes if I tried to use it in WSL2, even with the permissions and tried both winusb and libusb on the windows side of things.

knmcguire:
Due to our new monitoring system for github issues, I'll close this for now. I want to prevent PRs from being open. Also it is no longer mergable.

PR #229, opened 2021-04-23
syncCrazyFlie: remove prints

jonasdn:


krichardsson:
I think it is good to remove the prints but have a slight worry that it will be hard for the user to understand when/why a connection fails for instance. Should we use the logger instead to enable the user to choose what info to see?

jonasdn:
> I think it is good to remove the prints but have a slight worry that it will be hard for the user to understand when/why a connection fails for instance. Should we use the logger instead to enable the user to choose what info to see?

yes maybe! Added a commit to re-add them as `logger.debug`

jonasdn:
The prints are in the step-by-step tutorial (https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/user-guides/sbs_connect_log_param/) right now, so that would have to change. Do you have thoughts @knmcguire ?

PR #230, opened 2021-04-23
log: Fix checks for log config size and number of blocks and variables

jonasdn:


krichardsson (approved):
Looking good! 
This area has had multiple bugs in the past, seems as you found some more :-)

PR #233, opened 2021-05-06
Set cflinkcpp as default link driver on x86_64

jonasdn:
- Radiodriver: Set cflinkcpp as default
- setup.py: Require cflinkcpp on x86_64

Closes: #196

### DIFF
--- a/cflib/crtp/__init__.py
+++ b/cflib/crtp/__init__.py
@@ -26,7 +26,7 @@
 #  MA  02110-1301, USA.
 """Scans and creates communication interfaces."""
 import logging
-import os
+import platform
 
 from .exceptions import WrongUriType
 from .prrtdriver import PrrtDriver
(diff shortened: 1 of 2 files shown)


cflib/crtp/__init__.py
```diff
@@ -47,8 +47,11 @@
 def init_drivers(enable_debug_driver=False, enable_serial_driver=False):
     """Initialize all the drivers."""
 
-    env = os.getenv('USE_CFLINK')
```
ataffanel, line 50: What about keeling the env. variable to be able to keep experimenting with links?

If we keep it, I would like to have it trigger an error if the env. variable is set but the value does not correspond to a supported link (right now it defaults silently to python which can be confusing in case of spelling mistake)

jonasdn: Sounds nice, like the commit just added?

whoenig (commented):
This is awesome! Just one word of warning: we didn't extensively test cflinkcpp on Windows or Mac. I guess once we switch, we'll see (& fix) more issues.

jonasdn:
> This is awesome! Just one word of warning: we didn't extensively test cflinkcpp on Windows or Mac. I guess once we switch, we'll see (& fix) more issues.

Yes, awsomeness is coming. Right now we also cannot flash (at least warm boot) using this link, so there is some work left. But we will try to get it in shape for next release. And if not, we aim for next. Unstoppable force.

jonasdn:
We will revert if we get problems

PR #239, opened 2021-05-18
Made possible to give a velocity command and rate input at the same time for motion commander

knmcguire:
There has been a request to open up the _set_vel_setpoint to be able to give simultaneous velocity setpoint and yaw rate setpoint from the motion commander. 

https://forum.bitcraze.io/viewtopic.php?f=19&t=4855

### DIFF
--- a/cflib/positioning/motion_commander.py
+++ b/cflib/positioning/motion_commander.py
@@ -385,7 +385,7 @@ class MotionCommander:
 
     def start_linear_motion(self, velocity_x_m, velocity_y_m, velocity_z_m, rate_yaw=0.0):
         """
-        Start a linear motion. This function returns immediately.
+        Start a linear motion with an optional yaw rate input. This function returns immediately.
 
         positive X is forward
         positive Y is left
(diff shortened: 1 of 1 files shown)


knmcguire:
I just wondering if we need to add any tests for the new function eventhough it is just a wrapper. Also perhaps I can also add to the variable descriptions that the velocity commands are in body axis, as that is not super clear in my opinion.

krichardsson (approved):
Looking good!

A possible improvement could maybe be to rename `rate` to `rate_yaw` to be more explicit about what type of rate it is.

krichardsson:
An even better solution would maybe be to add a rate_yaw parameter with the default value 0.0 to the existing method start_linear_motion()?

`def start_linear_motion(self, velocity_x_m, velocity_y_m, velocity_z_m, rate_yaw=0.0):`

knmcguire:
true, that will be definitly easier. I'll change it now

knmcguire:
I just tested it in flight and seems to all work. Let's merge!

PR #241, opened 2021-06-02
Return a bool to indicate if a radio pk was sent

ataffanel:
This allows the caller to test directly if the packet was sent or if there was an error.


jonasdn (approved):
:+1:

PR #242, opened 2021-06-03
Allow nonzero landing height in PositionHlCommander

CoolBots:
We have encountered a need to land on objects (landing pads) that are situated at a different height than the takeoff (or, more specifically, calibration) height. The function `land()` in `crazyflie.HighLevelCommander` class accepts an `absolute_height_m` parameter, which is hard-coded to zero in `positioning.PositionHlCommander` class's version of this function (which is a pass-through function to the former one). This PR adds an optional parameter to the latter function, as well as introduces a method of specifying a `default_landing_height` at either initialization (`__init__` function), with default of 0.0, or as a separate function call to a newly introduced `set_landing_height` function. A very simple usage example was also added to `position_commander_demo.py` 

We find this functionality a useful addition, with no breaking changes to existing code. 

Unit test `test_that_it_goes_down_to_set_landing_height_on_landing` added; physically tested on Crazyflie 2.1 with Lighthouse deck.

### DIFF
--- a/cflib/positioning/position_hl_commander.py
+++ b/cflib/positioning/position_hl_commander.py
@@ -64,7 +64,7 @@ class PositionHlCommander:
         :param default_velocity: the default velocity to use
         :param default_height: the default height to fly at
         :param controller: Which underlying controller to use
-        :param default_landing_height: If taking off/landing on objects higher than ground, set this parameter to ensure proper landing
+        :param default_landing_height: Landing height (zero if not specified); for landing on objects off the ground
         """
         if isinstance(crazyflie, SyncCrazyflie):
             self._cf = crazyflie.cf
@@ -299,6 +299,6 @@ class PositionHlCommander:
     def set_landing_height(self, landing_height):
         """
         Set the landing height to a specific value
-        Use this function to land on objects that are at non-zero height                
+        Use this function to land on objects that are at non-zero height
         """
         self._default_landing_height = landing_height
--- a/cflib/positioning/position_hl_commander.py
+++ b/cflib/positioning/position_hl_commander.py
@@ -121,7 +121,7 @@ class PositionHlCommander:
         time.sleep(duration_s)
         self._z = height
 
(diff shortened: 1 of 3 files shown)


krichardsson:
Nice extension of the class, good idea!

I will review the code add some comments in the code.

There also seems to be some minor styling problems that break the build, the static code analyses tools we use are very (too?) picky. You can see the log here https://github.com/bitcraze/crazyflie-lib-python/runs/2735315626?check_suite_focus=false#step:3:100

Thanks!

cflib/positioning/position_hl_commander.py
```diff
@@ -117,7 +121,7 @@ def take_off(self, height=DEFAULT, velocity=DEFAULT):
         time.sleep(duration_s)
         self._z = height
 
-    def land(self, velocity=DEFAULT):
+    def land(self, velocity=DEFAULT, landing_height=0.0):
```
krichardsson, line 124: I think the user would expect a call to `land()` to use the `default_landing_height` that was set when the class was instantiated, the current implementation would try to land at 0.0. To get the desired functionality you will have to follow the same pattern that is used for the `velocity` parameter and set it to `DEFAULT`.

cflib/positioning/position_hl_commander.py
```diff
@@ -141,7 +145,7 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        self.land()
+        self.land(landing_height=self._default_landing_height)
```
krichardsson, line 148: With the proposed change to the `land()` method, `__exit__()` should only call `land()` without any arguments

krichardsson (changes requested):

CoolBots:
@krichardsson Very good point; I corrected my code and the affected unit test. I also fixed up the styling. Thanks!

krichardsson:
Merged.
Thanks!

PR #243, opened 2021-06-06
Make tutorial more user friendly

dominiknatter:
Here follow multiple minor updates to this step-by-step guide that increase user friendliness.

In "Step 2: Take off function", section "Changing the height":
In the first code snippet `mc.stop()` was not added. However, it is added in the full script (i.e., the code one can use to double check against before running it). As `mc.stop()` is called automatically when the motion commander instance is exited, I'd drop it for clarity.

In "Step 4: Logging while flying":
If one follows the snippets for a step-wise instruction, the code will not compile. This is because `lg_stab` should actually be called `logconf`.

In "Step 5: Combine logging and motion commander", section "Back and forth with limits":
In the first snippet a function `move_box_limit()` is introduced and the text suggests to place it above a non-existing function `move_in_box_limit()`. I updated it with `move_linear_simple()` in order to be consistent with the full ready-to-copy example.
The second snippet will not compile, because start_forward() is a function of the motion commander instance.
Furthermore, the three variables `body_x_cmd`, `body_y_cmd`, and `max_vel` are neither introduced nor needed in this section (only in the following one). Hence I'd drop them here for clarity.

### DIFF
--- a/docs/user-guides/sbs_motion_commander.md
+++ b/docs/user-guides/sbs_motion_commander.md
@@ -190,7 +190,6 @@ logging.basicConfig(level=logging.ERROR)
 def take_off_simple(scf):
     with MotionCommander(scf, default_height=DEFAULT_HEIGHT) as mc:
         time.sleep(3)
-        mc.stop()
 
 def param_deck_flow(name, value_str):
     ...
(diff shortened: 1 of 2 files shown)


knmcguire:
hi and thanks for this! Could you perhaps also update the scripts in examples/step-by-step ? those are the exact scripts from the tutorial

dominiknatter:
hey,
because most of my proposed updates were only in the code extracts (instead of the ready-to-copy code snippets), I only had to remove one line in the example code. Hope this works for you now.

knmcguire:
hi! I merged your PR. I had to add back the mc.stop() as the CI was giving an error that mc wasn't used in that function. 

Thanks!

PR #245, opened 2021-06-08
Make reset_to_firmware return bool

jonasdn:


cflib/bootloader/cloader.py
```diff
@@ -219,6 +219,7 @@ def reset_to_firmware(self, target_id):
                 break
 
         time.sleep(0.1)
+        return True
```
jonasdn, line 222: Can the loop above become infinite? @ataffanel ?

PR #247, opened 2021-06-17
cloader: Add retry timeout to update_info

jonasdn:
We have seen that this can fail in some paths, and we might need a retry
since this is talking to bootloader.


ataffanel (approved):
Making the communication function responsible for the retry is a good idea.

Looks good to me!

PR #250, opened 2021-06-29
bootloader: Close link after reset_to_firmware

jonasdn:
When we have successfully reset to firmware we should not keep the link
to the bootloader anymore. It confuses the state and causes bugs.

It will have start_bootloader() return True because it thinks the link
is already active.

Fixes: bitcraze/crazyflie-clients-python#520


ataffanel (approved):
Great work, this indeed fixes the bootloader bug!

Issue #155, opened 2020-06-10
Add additional description for package


evoggy:
When looking at this package at [pypi.org](https://pypi.org/project/cflib/) there's no longer description of the package. This should be added to the ```setup.py``` using the ```long_description``` and ```long_description_content_type``` tags. Either we write a new description or we use the readme.md file.

PR #254, opened 2021-09-06
Add additional decription for package

jonasdn:
Use our README.md to provide a long description for the cflib package
on pypi.org.

Closes: #155


ataffanel (approved):
Looks good! We should have done that a long time ago.

PR #259, opened 2021-09-14
Add rate limiter to Radio link

### LINKED ISSUE #257: Add P2P mode with relaxed CTRP timing
When connected to a Crazyflie, the lib will communicate as fast as possible with the Crazyflies. This means that there is no radio time left when using the P2P functionality implemented in bitcraze/crazyflie2-nrf-firmware#29 which prevents Crazyflie to talk to each-other the lib/client is connected to a Crazyflie.

One stop-gap solution, until we improve the radio link further, is to reduce the communication rate of the lib to free radio time for the P2P communication.

This could be done by adding one more optional parameter to the link: `radio://0/80/2M/E7E7E7e7E7?reduced_rate=10`. There is still at least 2 undecided questions:
 - What should the parameter name be?
 - Should the amount of bandwidth reduction be configurable?

ataffanel:
This will help with P2P connection by making the radio leave some air-time for P2P packets.

A new query "rate_limit=" is added to the radio URI. For example to limit the radio polling rate to 100Hz the uri would be: `radio://0/80/2M?rate_limit=100`

Fixes #257

### DIFF
--- a/cflib/crtp/radiodriver.py
+++ b/cflib/crtp/radiodriver.py
@@ -38,6 +38,7 @@ import queue
 import re
 import struct
 import threading
+import time
 from enum import Enum
 from queue import Queue
 from threading import Semaphore
@@ -46,8 +47,11 @@ from typing import Any
 from typing import Dict
 from typing import Iterable
 from typing import List
+from typing import Optional
 from typing import Tuple
 from typing import Union
+from urllib.parse import parse_qs
+from urllib.parse import urlparse
 
 import cflib.drivers.crazyradio as crazyradio
 from .crtpstack import CRTPPacket
@@ -258,7 +262,7 @@ class RadioDriver(CRTPDriver):
         an error message.
         """
 
-        devid, channel, datarate, address = self.parse_uri(uri)
+        devid, channel, datarate, address, rate_limit = self.parse_uri(uri)
         self.uri = uri
 
         if self._radio is None:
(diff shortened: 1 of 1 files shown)


cflib/crtp/radiodriver.py
```diff
@@ -641,6 +649,11 @@ def run(self):
                 else:
                     waitTime = 0
 
+            # If there is a rate limit setup, sleep here to force the rate limit
+            if self.rate_limit:
+                time.sleep(1.0/self.rate_limit)
```
jonasdn, line 654: Just a thought, nowhere in the code does anything state that `rate_limit` is in Hz ... should this be more explicit?

jonasdn (commented):
Nice!

Some questions:

- How does a user discover this feature?
  - (Where do we document it?)
 
- How do we tie this to p2p?
  - (How do people with p2p problem discover this solution?)

ataffanel:
Good questions!

> * How does a user discover this feature?
> * How do we tie this to p2p?

My though was to document it in the p2p page of the firmware: https://github.com/bitcraze/crazyflie-firmware/blob/master/docs/functional-areas/p2p_api.md, This should solve both questions.

jonasdn:
> Good questions!
> 
> > * How does a user discover this feature?
> > * How do we tie this to p2p?
> 
> My though was to document it in the p2p page of the firmware: https://github.com/bitcraze/crazyflie-firmware/blob/master/docs/functional-areas/p2p_api.md, This should solve both questions.

So this does only make sense for p2p? Should it then be explicit? `?php_mode=1` rather than `?rate_limit=rate`?  Or do we think that rate limit can be useful for other areas and maybe we  should document all query strings somewhere as well?

whoenig (approved):

whoenig:
Looks great - thanks! I also added https://github.com/bitcraze/crazyflie-link-cpp/issues/19, perhaps you want to create a similar one for the rust link.

@jonasdn: The idea of having a configurable limit is to allow a user-defined trade-off between p2p packets and host-computer packets. The rate_limit flag might also be useful in automated tests or benchmark to simulate a slow computer or lossy link.

cflib/crtp/radiodriver.py
ataffanel: An SI rate is normally Hz so I did not think about precising it. Are you envisioning a comment or renaming the variable with "_hz" in the name?

jonasdn: I am not sure, perhaps just some cognitive help while reading this part:

```Python
            # If there is a rate limit setup, sleep here to force the rate limit (Hz)
            if self.rate_limit:
                time.sleep(1.0/self.rate_limit)
```

jonasdn:
I added a PR to add documentation to the firmware here: https://github.com/bitcraze/crazyflie-firmware/pull/853

PR #269, opened 2021-10-08
Generate python documentation using pdoc3

jonasdn:
This commit generates API reference documentation using pdoc3.

The program pdoc3 will traverse the cflib module and generate
documentation based on the template file text.mako that we drop in the
template-dir.

The `text.mako` file is doing all the heavy lifting to create the
markdown we want.

For more information about pdoc3 and the custom template system see:
   https://pdoc3.github.io/pdoc/doc/pdoc/#custom-templates

---

---

### DIFF
--- a/tools/build-docs/build-docs
+++ b/tools/build-docs/build-docs
@@ -11,4 +11,17 @@ mkdir -p $scriptDir/tmp
 [ -n "$API_REF_BASE" ] || {
     export API_REF_BASE=$(readlink -fn $apiRefDir)
 }
+
+#
+# If cflib is not installed (mostly its dependencies) then pdoc3 cannot
+# traverse the library and get docstrings.
+#
+# The rewriting of HOME is a hack to make this work in a bare-bone Docker
+# environment.
+#
+pip3 show cflib > /dev/null 2>&1 || {  # this will run if cflib is not found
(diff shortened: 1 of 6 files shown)


krichardsson:
You must update module.json for the toolbelt to find the build-docs command.
It is a bit of a hack since we're not using doxygen, but it will work for now. (The correct solution would be to extend the toolbelt with "pdocs")

```
{
  "version": "2.0",
  "environmentReqs": {
    "build": ["python3"],
    "build-docs": ["doxygen"]
  }
}

```

ataffanel (approved):
Looks good!

tools/build-docs/build-docs
```diff
@@ -0,0 +1,22 @@
...
+[ -n "$API_REF_BASE" ] || {
+    export API_REF_BASE=$(readlink -fn $apiRefDir)
+}
+
+#
+# If cflib is not installed (mostly its dependencies) then pdoc3 cannot
+# traverse the library and get docstrings.
+#
+pip3 show cflib > /dev/null 2>&1 || {
+    pip3 install -e $scriptDir/../../
```
krichardsson, line 20: A bit of a side effect if someone runs the script locally but makes sense in the web-builder container :-) I think this is good enough.

PR #273, opened 2021-10-22
toc: Fix TocFetcher requesting ToC entries when none are available

WowSuchName:
I noticed that the lib requests ToC entries from the Crazyflie when the latter returns that there are none. This should fix it without breaking the existing log messages.


knmcguire:
Sorry for the late repy! It looks alright to me but I want to have some others to take a look at this as well to make sure this goes all according to it's purpose. I'll assign some people.

whoenig (approved):
Not an expert, but looks sound to me as well.

knmcguire:
alright lets merge it!

PR #278, opened 2021-11-02
Convert six statements to the usage of augmented assignments

elfring:
[Augmented assignment statements became available with Python 2](https://docs.python.org/3/whatsnew/2.0.html#augmented-assignment "Augmented assignments").
:thought_balloon: Thus improve six source code places accordingly.


krichardsson:
For some reason, this has not been visible to us until now. Anyway it is now merged. Thanks!
closes #277

elfring:
>For some reason, this has not been visible to us until now.

I can share another bit of background information if you would like to know it.


>Anyway it is now merged.

Thanks for your change acceptance (after some delay).

PR #280, opened 2021-11-11
Add initial persistent parameter API

jonasdn:
A persistent parameter is one that _can_ be stored to eeprom. The way to do this is to set a value to the parameter, using the regular parameter API, and then use CRTP to tell the firmware to store the value. A value can also be cleared from eeprom using CRTP.

In order to build nice UIs we want to be able to tell whether a persistent parameter is stored or not, and what the default value of the parameter is. This is done by asking the firmware about the state.

Also! Since we are taking the last bit in the parameter type to express that this is a persistent parameter we have now introduced the concept of extended type information. That we also fetch as part of the connection in the Crazyflie class.

Contains the following commit:
- Param: fetch extended types at connection
- Param: add persistent_get_state method
- Param: add persistent_store method
- Param: add persistent_clear method

### DIFF
--- a/cflib/crazyflie/param.py
+++ b/cflib/crazyflie/param.py
@@ -437,9 +437,11 @@ class Param():
 
                 is_stored = pk.data[3] == 1
                 if not is_stored:
-                    default_value = struct.unpack(f'<{element.pytype}')
+                    default_value, = struct.unpack(element.pytype, pk.data[4:])
                 else:
-                    default_value, stored_value = struct.unpack(f'<{element.pytype}*2')
+                    # Remove little-endian indicator ('<')
+                    just_type = element.pytype[1:]
+                    default_value, stored_value = struct.unpack(f'<{just_type * 2}', pk.data[4:])
 
                 callback(complete_name,
                          PersistentParamState(
--- a/cflib/crazyflie/param.py
+++ b/cflib/crazyflie/param.py
@@ -446,7 +446,7 @@ class Param():
                 else:
                     # Remove little-endian indicator ('<')
                     just_type = element.pytype[1:]
-                    stored_value, default_value = struct.unpack(f'<{just_type * 2}', pk.data[4:])
+                    default_value, stored_value = struct.unpack(f'<{just_type * 2}', pk.data[4:])
 
                 callback(complete_name,
                          PersistentParamState(
--- a/cflib/crazyflie/param.py
+++ b/cflib/crazyflie/param.py
@@ -70,6 +70,7 @@ PersistentParamState = namedtuple('PersistentParamState', 'is_stored default_val
 
 MISC_PERSISTENT_STORE = 3
 MISC_PERSISTENT_GET_STATE = 4
+MISC_PERSISTENT_CLEAR = 5
 
 # One element entry in the TOC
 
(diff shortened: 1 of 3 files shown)


cflib/crazyflie/param.py
```diff
@@ -313,6 +358,166 @@ def get_value(self, complete_name, timeout=60):
...
+        argument on success and with `False` as an argument on failure.
+
+        @param complete_name The 'group.name' name of the parameter to store
+        @param callback Optional callback should take boolean status as arg
+        """
+        element = self.toc.get_element_by_complete_name(complete_name)
+
+        def new_packet_cb(pk):
+            if pk.channel == MISC_CHANNEL and pk.data[0] == MISC_PERSISTENT_CLEAR:
+                callback(pk.data[3] == 0)
```
krichardsson, line 374: `callback` is optional and possibly None, I can not see any check that it actually contains a function before it is called. Maybe I miss something but otherwise a `not None` check should be added somewhere.

cflib/crazyflie/param.py
```diff
@@ -313,6 +358,166 @@ def get_value(self, complete_name, timeout=60):
...
+        argument on success, and with `False` as an argument on failure.
+
+        @param complete_name The 'group.name' name of the parameter to store
+        @param callback Optional callback should take boolean status as arg
+        """
+        element = self.toc.get_element_by_complete_name(complete_name)
+
+        def new_packet_cb(pk):
+            if pk.channel == MISC_CHANNEL and pk.data[0] == MISC_PERSISTENT_STORE:
+                callback(pk.data[3] == 0)
```
krichardsson, line 395: `callback` can be None, se above

krichardsson (changes requested):
I just had a quick look

cflib/crazyflie/param.py
jonasdn: Yes, nice catch! Thanks!

cflib/crazyflie/param.py
jonasdn: Thanks!

cflib/crazyflie/param.py
jonasdn: I added a check for callback and forced push. Also make sure to remove port callbacks now.

jonasdn:
> I just had a quick look

Does it look better now?

examples/parameters/persistent_params.py
```diff
@@ -0,0 +1,96 @@
...
+
+        # Collect the names of all persistent parameters
+        persistent_params = []
+        for group_name, params in scf.cf.param.toc.toc.items():
+            for param_name, element in params.items():
+                if element.is_persistent():
+                    complete_name = group_name + '.' + param_name
+                    persistent_params.append(complete_name)
+
+        if len(persistent_params) == 0:
```
jonasdn, line 62: nit: this _could_ be:
```python
if not persistent_params:
```

Since a list is `True` if non-empty.

examples/parameters/persistent_params.py
```diff
@@ -0,0 +1,96 @@
...
+            sys.exit(0)
+
+        # Get state for the persistent parameters
+        print('-- All persistent parameters --')
+
+        def state_callback(complete_name, state):
+            print(complete_name, state)
+
+            persistent_params.pop(0)
+            if len(persistent_params) > 0:
```
jonasdn, line 73: same here _could_ be:
```python
if persistent_params:
```

examples/parameters/persistent_params.py
```diff
@@ -0,0 +1,96 @@
...
+
+        if len(persistent_params) == 0:
+            print('No persistent parameters')
+            sys.exit(0)
+
+        # Get state for the persistent parameters
+        print('-- All persistent parameters --')
+
+        def state_callback(complete_name, state):
+            print(complete_name, state)
```
jonasdn, line 70: perhaps:

```python
print(f'{complete_name}: {state}')
```

?

examples/parameters/persistent_params.py
```diff
@@ -0,0 +1,96 @@
...
+        scf.cf.param.persistent_get_state(persistent_params[0], state_callback)
+        # Wait for all parameters to be done
+        wait_for_callback_event.wait()
+
+        print()
+        print('-- Set the default LED ring effect --')
+        led_ring_param_name = 'ring.effect'
+        scf.cf.param.set_value(led_ring_param_name, 7)
+
+        def is_stored_callback():
```
jonasdn, line 90: This should take arguments that indicate success, something like:

```python
def is_stored_callback(complete_name, success):
  if success:
    print(f'Persisted {complete_name}!')
  else:
   print(f'Failed to persist {complete_name}!')
```
?

jonasdn (commented):
Looks good! Some comments below!

PR #284, opened 2021-12-02
param: Add blocking timeout of max 1s to set_value

jonasdn:
Wait for reply from cf for up to a second after setting a parameter with
`param.setvalue()`


jonasdn:
Without this you would need to add a sleep for instance between `param.set_value()` and `param.persistent_store()`

jonasdn:
Choose other method

PR #292, opened 2021-12-20
param: Add get_default_value method

jonasdn:


cflib/crazyflie/param.py
```diff
@@ -358,6 +359,36 @@ def get_value(self, complete_name, timeout=60):
         [group, name] = complete_name.split('.')
         return self.values[group][name]
 
+    def get_default_value(self, complete_name, callback=None):
+        """
+        Clear the current value of the specified parameter. from
```
krichardsson, line 364: The description seems to be copy/pasted from the clear function :-)

krichardsson (changes requested):
Looks good to me (apart from the description)

cflib/crazyflie/param.py
jonasdn: Yes :) Some kind of frankenstein state, have fixed with force push

PR #296, opened 2021-12-23
Examples: Add console.py

jonasdn:
This adds a simple example of using the console class to receive
DEBUG_PRINTS from the Crazyflie.

Sample output:
 ```
 $ CFLIB_URI=radio://0/10/2M/DEADBEEF python3 examples/console/console.py
SYS: ----------------------------
SYS: Crazyflie 2.1 is up and running!
SYS: Production release 2021.06
SYS: I am 0x203937434848500D00470039 and I have 1024KB of flash!
CFGBLK: v1, verification [OK]
DECK_CORE: 1 deck(s) found
DECK_CORE: Calling INIT on driver bcFlow2 for deck 0
ZR2: Z-down sensor [OK]
PMW: Motion chip id: 0x49:0xB6
IMU: BMI088: Using I2C interface.
IMU: BMI088 Gyro connection [OK].
IMU: BMI088 Accel connection [OK]
IMU: BMP388 I2C connection [OK]
ESTIMATOR: Using Kalman (2) estimator
CONTROLLER: Using PID (1) controller
MTR-DRV: Using brushed motor driver
SYS: About to run tests in system.c.
EEPROM: I2C connection [OK].
STORAGE: Storage check [OK].
IMU: BMI088 gyro self-test [OK]
DECK_CORE: Deck 0 test [OK].
SYS: Self test passed!
STAB: Wait for sensor calibration...
SYS: Free heap: 17672 bytes
[host] Connected, use ctrl-c to quit.
```

### DIFF
--- a/examples/console/console.py
+++ b/examples/console/console.py
@@ -1,0 +1,61 @@
+#
+#     ||          ____  _ __
+#  +------+      / __ )(_) /_______________ _____  ___
+#  | 0xBC |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+#  +------+    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+#   ||  ||    /_____/_/\__/\___/_/   \__,_/ /___/\___/
+#
+#  Copyright (C) 2021 Bitcraze AB
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+import time
+
+import cflib.crtp  # noqa
+from cflib.crazyflie import Crazyflie
+from cflib.crazyflie.syncCrazyflie import SyncCrazyflie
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by knmcguire — APPROVED** [review]
I tried it out and it works good! 

perhaps you can think about removing the parameter examples to keep it more minimal example form.

**Comment by jonasdn** [comment]
> I tried it out and it works good!
> 
> perhaps you can think about removing the parameter examples to keep it more minimal example form.

I was setting the params to foce output on console, otherwise if you run this against a cf that has been running for a while, you get no input and it is hard to gather if it is working. But maybe that is overthinking it?

**Comment by knmcguire** [comment]
ah yeah true, but that is also something that we can add to the explanation text above or the readout for the ctrl-c ?

Crazyswarm's console program is only the console output

**Comment by jonasdn** [comment]
> ah yeah true, but that is also something that we can add to the explanation text above or the readout for the ctrl-c text ?
> 
> Crazyswarm's console program is only the console output

It is in the output:

```
$ CFLIB_URI=radio://0/10/2M/DEADBEEF python3 examples/console/console.py
[host] Connected, use ctrl-c to quit.
[host] Changing controller, to force console output
[host] Setting Mellinger controller
CONTROLLER: Using Mellinger (2) controller
[host] Setting PID controller
CONTROLLER: Using PID (1) controller
```

**Comment by knmcguire** [comment]
ah that is same as the sample output and what I saw. Perhaps forgot to copy it?

I was thinking more of, '[host] If full output of the console is not visible, try to run this script right after CF startup'

You can also add it as a comment on top of the script.

```
"""
Simple console output example that connects to one crazyflie and prints out the logging information from the firmware.  

If full output of the console is not visible, try to run this script right after CF startup
"""
```

It might be good to add at least the top line anyway on top of the script, as all the other examples have an short brief explanation as well.

jonasdn:
Agreed, updated.

PR #297, opened 2022-01-09
Full lighthouse base station geometry estimation

krichardsson:
This pull request adds improved base station geometry estimation to the lighthouse system. It supports more than 2 base stations (if supported by the firmware) as well as better alignment of the coordinate system. A side effect of the process is also a measurement of the errors in the solution (how good the estimated geometry is).

The new solution requires measurements with a Crazyflie in the flight space and a script has been added to the examples (examples/lighthouse/bs_geometry_estimation.py), to do a full geometry estimation and upload it to a Crazyflie. The script displays instructions and will hopefully be useful for now. The client should be updated to support the full process, until then the new implementation is compatible with the current client (but with limited functionality).

Open CV is not needed any more and the dependency has been removed. A new dependency to scipy has been added instead.

There are some issues that need to be solved to make the code work with the build tools: 
* I think I used type hint features that were introduced in python 3.8 and we currently support older versions. How to handle this?
* I use numpy.typing which currently is not supported by the builder docker image.

The PR is pretty massive, let's discuss if we want to split it up

### DIFF
--- a/cflib/localization/ippe_cf.py
+++ b/cflib/localization/ippe_cf.py
@@ -22,6 +22,7 @@
 from collections import namedtuple
 
 import numpy as np
+import numpy.typing as npt
 
 from ._ippe import mat_run
 
@@ -46,7 +47,7 @@ class IppeCf:
     Solution = namedtuple('Solution', ['R', 't', 'reproj_err'])
 
     @staticmethod
-    def mat_run(U_cf, Q_cf):
+    def solve(U_cf: npt.ArrayLike, Q_cf: npt.ArrayLike) -> list[Solution]:
         """
         The solution to Perspective IPPE with point correspondences computed
         between points in world coordinates on the plane z=0, and normalised points in the
--- a/cflib/localization/ippe_cf.py
+++ b/cflib/localization/ippe_cf.py
@@ -57,24 +57,17 @@ class IppeCf:
 
         This is a wrapper function to convert from/to CF coordinate system/array style
 
-        Parameters
-        ----------
-        U_cf: Nx3 matrix holding the model points in world coordinates.
+        :param U_cf: Nx3 matrix holding the model points in world coordinates.
+        :param Q_cf: Nx2 matrix holding the points in the image. These are in normalised
+                     pixel coordinates. That is, the effects of the camera's intrinsic matrix
+                     and lens distortion are corrected, so that the Q projects with a perfect
+                     pinhole model.
 
-        Q_cf: Nx2 matrix holding the points in the image. These are in normalised
-            pixel coordinates. That is, the effects of the camera's intrinsic matrix
-            and lens distortion are corrected, so that the Q projects with a perfect
-            pinhole model.
-
-            First param: Y (positive to the left)
-            Second param: Z (positive up)
-
-
-        Return
-        ---------
-        IPPEPoses: A list that contains 2 sets of pose solution from IPPE including rotation matrix
-            translation matrix, and reprojection error. The first solution in the list has
-            the smallest reprojection error.
+                     First param: Y (positive to the left)
+                     Second param: Z (positive up)
+        :return: A list that contains 2 sets of pose solution from IPPE including rotation matrix
+                 translation matrix, and reprojection error. The first solution in the list has
+                 the smallest reprojection error.
         """
 
         U, Q = IppeCf._cf_to_ippe(U_cf, Q_cf)
--- a/cflib/localization/ippe_cf.py
+++ b/cflib/localization/ippe_cf.py
@@ -19,6 +19,8 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with this program. If not, see <http://www.gnu.org/licenses/>.
+from __future__ import annotations
+
 from collections import namedtuple
 
 import numpy as np
--- a/cflib/localization/ippe_cf.py
+++ b/cflib/localization/ippe_cf.py
@@ -1,0 +1,121 @@
+# -*- coding: utf-8 -*-
+#
+# ,---------,       ____  _ __
+# |  ,-^-,  |      / __ )(_) /_______________ _____  ___
+# | (  O  ) |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+# | / ,--'  |    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+#    +------`   /_____/_/\__/\___/_/   \__,_/ /___/\___/
+#
+# Copyright (C) 2022 Bitcraze AB
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, in version 3.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program. If not, see <http://www.gnu.org/licenses/>.
+from collections import namedtuple
+
+import numpy as np
+
+from ._ippe import mat_run
+
+
+class IppeCf:
+    """
+    A wrapper class to simplify usage of IPPE in CF code.
+    Converts between CF style of coordinate systems/data structures and
+    IPPE (Open CV) style.
+    """
+
+    # Rotation matrix to transform from IPPE to CF
+    _R_ippe_to_cf = np.array([
+        [0.0, 0.0, 1.0],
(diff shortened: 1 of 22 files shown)


krichardsson:
Seems to be a problem in the alignment step, will look into it

cflib/localization/ippe_cf.py
```diff
@@ -0,0 +1,122 @@
...
+        [-1.0, 0.0, 0.0],
+        [0.0, -1.0, 0.0],
+    ])
+
+    # Rotation matrix to transform from CF to IPPE
+    _R_cf_to_ippe = np.transpose(_R_ippe_to_cf)
+
+    Solution = namedtuple('Solution', ['R', 't', 'reproj_err'])
+
+    @staticmethod
```
jonasdn, line 49: Nice with types! Maybe we should use the renderable parameter description style in docstring though?

```
:param U_cf: Nx3 matrix holding the model points in world coordinates.
:param Q_cf: Nx2 matrix holding the points in the image. These are in normalised
            pixel coordinates. That is, the effects of the camera's intrinsic matrix
            and lens distortion are corrected, so that the Q projects with a perfect
            pinhole model.

:return: A list that contains 2 sets of pose solution from IPPE including rotation matrix
            translation matrix, and reprojection error. The first solution in the list has
            the smallest reprojection error.
```

cflib/localization/lighthouse_bs_geo.py
```diff
@@ -35,48 +37,10 @@ class LighthouseBsGeoEstimator:
...
-                [1.0, 0.0, 0.0],
-                [0.0, 1.0, 0.0],
-                [0.0, 0.0, 1.0]
-            ])
-
-        self._dist_coef = np.zeros(4)
-
         # Sanity check maximum pos
-        self._sanity_max_pos = 10
+        self._sanity_max_pos = 15
```
jonasdn, line 41: what is this sanity value based on?

cflib/localization/lighthouse_bs_vector.py
```diff
@@ -22,20 +22,21 @@
...
-    two light planes.
+    two light planes defined by angles measured by a base station.
     It also provides functionality to convert between lighthouse V1 angles,
     V2 angles and cartesian coordinates.
     """
 
     T = math.pi / 6
 
-    def __init__(self, lh_v1_horiz_angle, lh_v1_vert_angle):
+    def __init__(self, lh_v1_horiz_angle: float, lh_v1_vert_angle: float) -> None:
```
jonasdn, line 39: nice type additions!

cflib/localization/lighthouse_bs_vector.py
```diff
@@ -70,41 +71,93 @@ def from_cart(cls, cart_vector):
...
+        return np.float32((math.tan(self._lh_v1_horiz_angle), math.tan(self._lh_v1_vert_angle)))
+
     def _q(self):
         return math.tan(self._lh_v1_vert_angle) / math.sqrt(1 + math.tan(self._lh_v1_horiz_angle) ** 2)
+
+
+# A LighthouseBsVectors is a list of 4 LighthouseBsVector, one for
+# each sensor
+# LighthouseBsVectors = list[LighthouseBsVector]
+
```
jonasdn, line 142: Perhaps this could be a class docstring? So it shows up in rendered docs?

cflib/localization/lighthouse_geometry_solver.py
```diff
@@ -0,0 +1,429 @@
...
+
+    An examples matrix:
+
+                        bs0_pose, bs1_pose, bs2_pose, bs3_pose, cf1_pose, cf2_pose, ...
+
+    cf0/bs2/sens0/ang0                        X
+    cf0/bs2/sens0/ang1                        X
+    cf0/bs2/sens1/ang0                        X
+    cf0/bs2/sens1/ang1                        X
+    ...
```
jonasdn, line 84: Great comment!

cflib/localization/lighthouse_geometry_solver.py
```diff
@@ -0,0 +1,429 @@
...
+        """
+        To speed up calculations all operations in the iteration phase are done on np.arrays of equal length (ish),
+        the numpy flavour of parallell work. Some data is reused in multiple equations (for instance sensor
+        positions) and to avoid copying of data we use np.arrays as indexes. This method creates the necessary
+        indexes.
+
+        Since the equation system we are solving is sparse, we can also do a second optimization and only calculate
+        the parts of the matrix that are non-zero. We have to provide a matrix containing 1 in the positions where
+        the jacobian is non-zero, this matrix is also generated here.
+        """
```
jonasdn, line 186: nice comment!

cflib/localization/lighthouse_initial_estimator.py
```diff
@@ -0,0 +1,166 @@
...
+        """
+        Make a rough estimate of the poses of all base stations and CF poses found in the samples.
+
+        The pose of the Crazyflie in the first sample is used as a reference and will define the
+        global reference frame.
+
+        :param matched_samples: A list of samples with lihghthouse angles.
+        :param sensor_positions: An array with the sensor positions on the lighthouse deck (3D, CF ref frame)
+        :return: a
+        """
```
jonasdn, line 49: :+1:

cflib/localization/lighthouse_initial_estimator.py
```diff
@@ -0,0 +1,166 @@
...
+
+        bs_poses_ref_cfs = cls._angles_to_poses(matched_samples, sensor_positions)
+
+        # Use the first CF pose as the global reference frame
+        bs_poses: dict[int, Pose] = bs_poses_ref_cfs[0]
+        # TODO Do not use first sample as reference, pass it in as a parameter
```
jonasdn, line 55: Should this TODO be an issue? When will we ever get around to do this? Could it be done as part of this PR? Could the TODO be removed?

TODO in code is sort of a pet peave of mine, they usually stay forever :)

examples/lighthouse/bs_geometry_estimation.py
```diff
@@ -0,0 +1,213 @@
...
+#  of the License, or (at your option) any later version.
+#
+#  This program is distributed in the hope that it will be useful,
+#  but WITHOUT ANY WARRANTY; without even the implied warranty of
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#  You should have received a copy of the GNU General Public License
+#  along with this program; if not, write to the Free Software
+#  Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston,
+#  MA  02110-1301, USA.
```
jonasdn, line 23: Please use the same blurb in other files, we do not want to keep track of the office of the Free Software Foundation. And  the recommended blurb is the one you used prior:

```
This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.

This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.

You should have received a copy of the GNU General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
```
https://www.gnu.org/licenses/gpl-howto.en.html

examples/lighthouse/bs_geometry_estimation.py
```diff
@@ -0,0 +1,213 @@
...
+the instructions.
+
+This script supports large systems with more than 2 base stations (if
+supported by the CF firmware).
+
+This script is a temporary implementation until similar functionality has been
+added to the client.
+
+Prerequisite: The base station calibration data must have been
+received by the Crazyflie before this script is executed.
```
jonasdn, line 40: Great with an example!

jonasdn (commented):
Really nice and clean code @krichardsson !

Yes it is complex, but you have worked hard! Both with solving the problem and structuring the code! I love the type conversions, and the detailed comments, with parameter specification! 

Your effort shows!

krichardsson:
It turned out that the alignment was not the problem, but the initial guess was bad. Fixed now.

cflib/localization/ippe_cf.py
krichardsson: Agree! It was reused from the original ippe, my bad

cflib/localization/lighthouse_geometry_solver.py
krichardsson: Thanks!

examples/lighthouse/bs_geometry_estimation.py
krichardsson: Fixed!

jonasdn (approved):
:+1:

PR #299, opened 2022-01-11
Use Pypi-packaged libusb library instead of the OS installed one

ataffanel:
This PR makes use of the pypi [libus-package](https://pypi.org/project/libusb-package/) pip package to pull the libusb dll/so library. This allows the lib
to communicate with the Crazyradio and Crazyflie without requiring extra install in the system.

This should simplify the installation and usage of the lib in MacOS and should allow to use the winUSB driver on Windows.

This is tested on Linux/amd64. It needs to be tested on:
 - [ ] Windows
 - [ ] MacOS x86_64
 - [ ] MacOX arm64
 - [ ] Linux arm64?


cflib/drivers/cfusb.py
```diff
@@ -29,6 +29,8 @@
 import os
```
jonasdn, line 29: `os` is not used anymore?

cflib/drivers/crazyradio.py
```diff
@@ -30,6 +30,8 @@
 import platform
 
 import usb
```
jonasdn, line 32: Unused imports

jonasdn (commented):
Nice deletions!

cflib/drivers/cfusb.py
ataffanel: Is is now re-used thanks to handling special case for the special Windows ...

PR #302, opened 2022-01-20
Bump minimum Python version to 3.7

ataffanel:
This is required because of dependencies requirering Python 3.7.

Python 3.6 is not officialy supported by python.org anymore.

This breaks support for stock Ubuntu 18.04 which should not be a big issue since we are so far in the current Ubuntu LTS 20.04 and very close to the next 22.04.

Documentation has been updated to contain the minimum supported python version.

### DIFF
--- a/docs/installation/install.md
+++ b/docs/installation/install.md
@@ -3,6 +3,12 @@ title: Installation
 page_id: install
 ---
 
+## Requirements
+
+This project requires Python 3.7+.
+For Ubuntu it means that Ubuntu 20.04+ is supported unless a more recent version of Python is installed.
+
+See bellow sections for more platform-specific requirements.
 ## Development
 ### Developing for the cflib
 * [Fork the cflib](https://help.github.com/articles/fork-a-repo/)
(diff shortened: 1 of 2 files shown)


docs/installation/install.md
```diff
@@ -3,6 +3,12 @@ title: Installation
 page_id: install
 ---
 
+## Requirements
+
+This project requires Python 3.7+.
+For Ubuntu it means that Ubuntu 20.04+ is supported unless a more recent version of Python is installed.
```
knmcguire, line 9: This sentence is a bit too easy to misinterpreted. Perhaps just remove so that people can find the version for them self?

docs/installation/install.md
```diff
@@ -3,6 +3,12 @@ title: Installation
...
+
+See bellow sections for more platform-specific requirements.
```
knmcguire, line 11: typo 'below'

docs/installation/install.md
knmcguire: Perhaps better to say. Ubuntu 20.04 has this installed for standard but for 18.04 this will need to be upgraded? and then also say that anything below 18.04 is not supported

ataffanel: Removed. Ubuntu 16.04 is not supported for free by Canonical anymore so I just ignored it (it was already out of support from us).

PR #305, opened 2022-01-24
fixed parsing xml qtm

knmcguire:
Same fix as the client qualysis tab fix https://github.com/bitcraze/crazyflie-clients-python/pull/567  but now with the qualysis example script


knmcguire:
there is no qtm version of setup.py here, so no need to bump it up like we did for the cfclient

PR #312, opened 2022-01-25
Fix USB connection to CF

### LINKED ISSUE #264: USB connection to CF not always set up properly
Repeated connections to a Crazyflie via USB seems to sometimes hang in the connection phase. 

How to reproduce 1:
* Open the python client
* Connect CF to USB, scan in client and choose usb://0
* Click "Connect"
* Expected behavior: the button should change to "Cancel" and after a while to "Disconnect" to indicate the connection process is complete
* Observed behavior: sometimes the button changes to "Cancel" only and the process seems to stop
* Click "Disconnect" (or "Cancel" in case of failed connection)
* Repeat Connection process

How to reproduce 2:
* Write a python script that connects via USB, for instance using the SyncCrazyflie class
* Run the script multiple times
* Sometimes the connection process hangs

Found on:
On Ubuntu 21.04
Firmware 2021.06

ataffanel:
Adding a USB reset seems to fix the USB connection to the Crazyflie when
trying to connect multiple times.

This has been tested on one Linux machine. This needs to be tested on
other machines and OS. Follow the procedure explained on the ticket #264 to test.

Fixes #264

### DIFF
--- a/cflib/crtp/usbdriver.py
+++ b/cflib/crtp/usbdriver.py
@@ -70,16 +70,15 @@ class UsbDriver(CRTPDriver):
         """
 
         # check if the URI is a radio URI
-        if not re.search('^usb://', uri):
+        uri_data = re.search('^usb://([0-9]+)$',
+                             uri)
+        if not uri_data:
             raise WrongUriType('Not a radio URI')
 
         # Open the USB dongle
         if not re.search('^usb://([0-9]+)$',
                          uri):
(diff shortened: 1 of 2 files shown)


jonasdn (commented):
Fixes the problem on my machine :1st_place_medal:

krichardsson (approved):

krichardsson:
Ok except pep8 problems

ataffanel:
Still works fine on Windows and MacOS

Issue #314, opened 2022-01-26
Errors in examples


knmcguire:
Hi!

There are some examples that are currently failing:
- sbs-motion-commander: [this line](https://github.com/bitcraze/crazyflie-lib-python/blob/715953829194322e45540a8fc93c67ac415a80a8/examples/step-by-step/sbs_motion_commander.py#L122-L124) need a pause between logging initialization and the ifstatement that uses the variable from it
- PID_controller_tunner: [This line](https://github.com/bitcraze/crazyflie-lib-python/blob/715953829194322e45540a8fc93c67ac415a80a8/examples/tuning/PID_controller_tuner.py#L154-L156) is polling for a parameter that does not exist anymore (posCtlPid.xyVelMax), these are seperated now in xVelMax and yVelMax (see the log doc [here](https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/api/params/#posctlpid))

PR #315, opened 2022-01-26
Fix errors in examples

jonasdn:
- sbs_motion_commander.py: Use Event instead of boolean
- sbs_motion_commander.md: Update to keep in sync
- PID_controller_tuner: handle velMax

Closes: #314

### DIFF
--- a/examples/tuning/PID_controller_tuner.py
+++ b/examples/tuning/PID_controller_tuner.py
@@ -152,7 +152,10 @@ class TunerControlCF:
         self.cf.param.add_update_callback(
             group='posCtlPid', name='zKd', cb=self.param_updated_callback_Kd)
         self.cf.param.add_update_callback(
-            group='posCtlPid', name='xyVelMax',
+            group='posCtlPid', name='xVelMax',
+            cb=self.param_updated_callback_vMax)
+        self.cf.param.add_update_callback(
+            group='posCtlPid', name='yVelMax',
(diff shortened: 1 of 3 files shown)


jonasdn:
An alternate approach to the PID_controller_tuner is to delete it for now, and revisit tuning script/GUI in next release. Hmm, @knmcguire ?

knmcguire:
True, but let's keep it for now. It's featured in one of my EPFL lectures.

examples/tuning/PID_controller_tuner.py
```diff
@@ -152,7 +152,10 @@ def __init__(self, pid_gui, scf):
         self.cf.param.add_update_callback(
             group='posCtlPid', name='zKd', cb=self.param_updated_callback_Kd)
         self.cf.param.add_update_callback(
-            group='posCtlPid', name='xyVelMax',
+            group='posCtlPid', name='xVelMax',
```
knmcguire, line 155: Actually this will still fail as the posCtlPid.xyVelMax parameter is being used in multiple places, when requesting the parameter and updating the value. So those need to be changed as well.

knmcguire (changes requested):

examples/tuning/PID_controller_tuner.py
jonasdn: doh! thanks!

Pushed a new commit

knmcguire (approved):

PR #316, opened 2022-01-26
build-docs: set -e to detect errors

jonasdn:


jonasdn:
Without this the website build will not detect errors with generating cflib docs

krichardsson:
Good catch!

PR #324, opened 2022-03-11
Improved multi bs estimation

krichardsson:
* Reworked initial guess functionality that is better at figuring out which solution from ippe that is the correct one. 
* This version does not have any apecial limitatoins on CF or base station poses, but requires 2 or more base stations to work.


knmcguire:
I think it looks good! I wanted to double check the build-docs for the api generated docs, but getting an error for that now from the toolbelt.

krichardsson:
Build-docs works on my machine. What error do you get?

knmcguire:
`ImportError: Error importing 'cflib.crazyflie': ModuleNotFoundError: No module named 'libusb_package'`

PR #327, opened 2022-03-23
Add suport for flashing both ESP and GAP8 on the AI-deck

krichardsson:
Support updated deck memory mapping and more complex flashing

* Updated deck memory mapping. New protocol version, requires bitcraze/crazyflie-firmware#1003
* Support for flashing multiple decks from one zip, with reboot between each deck


evoggy (approved):
Looks good!

PR #334, opened 2022-05-10
example fixes

knmcguire:
Here are some tiny improvements to the example scripts for the cflib


examples/parameters/basicparam.py
```diff
@@ -125,6 +125,8 @@ def _param_callback(self, name, value):
             self._cf.param.set_value('pid_attitude.pitch_kd',
                                      '{:.2f}'.format(pkd))
 
+            self._cf.param.request_param_update('pid_attitude.pitch_kd')
```
krichardsson, line 128: I think we should not add this line as it is not required, but possibly adds to the confusion

krichardsson (changes requested):
Looking good, nice clean up!

PR #336, opened 2022-05-11
fix to cflib doc

knmcguire:
There were quite some outdated instructions in the install documentation. I also removed a line that was a bit too optimistic for the matlab installation instructions.


evoggy (approved):
Looks good!

PR #342, opened 2022-06-23
Add CPX to the library

evoggy:
This change adds CPX to the library. It also has the following updates:

* Modify Serial driver to use CPX instead of syslink

This requires the crazyflie firmware to be updated as well with the following PR: https://github.com/bitcraze/crazyflie-firmware/pull/1068


cflib/crtp/serialdriver.py
```diff
@@ -99,25 +90,29 @@ def connect(self, uri, linkQualityCallback, linkErrorCallback):
...
+        # Switch the link bridge to CPX in the Crazyflie
+        self.cpx.sendPacket(CPXPacket(destination=CPXTarget.STM32,
+                                      function=CPXFunction.SYSTEM,
+                                      data=[0x21, 0x01]))
+        # Force client connect to true
+        self.cpx.sendPacket(CPXPacket(destination=CPXTarget.STM32,
+                                      function=CPXFunction.SYSTEM,
+                                      data=[0x20, 0x01]))
+
+        # TODO! These should be reset again!
```
knmcguire, line 109: Make an issue for this

knmcguire:
How about unit tests? Or rather add it to the test lab because it is all communication based

cflib/crtp/serialdriver.py
evoggy: With the intended usage this is not really needed (i.e connecting with a Raspberry Pi onboard)

knmcguire (approved):

PR #352, opened 2022-08-08
Step by Step Swarm Guide

marios-stam:
A step-by-step guide for the Swarm class is introduced.

Could you provide feedback and other things that could be included as well?

### DIFF
--- a/docs/user-guides/sbs_swarm_interface.md
+++ b/docs/user-guides/sbs_swarm_interface.md
@@ -3,7 +3,7 @@ title: "Step-by-Step: Swarm Interface"
 page_id: sbs_swarm_interface
 ---
 
-Here we will go through step-by-step how to interface with a swarm of crazyflies and make all the copters of the swarm hover and fly simultaneously in a square shape using the `Swarm()` class of the cflib.For this tutorial you will need a swarm (2 or more) of crazyflies with the latest firmware version installed and a positioning system (Flowdeck,Lighthouse,LPS etc.) that is able to provide data for the position estimation of the crazyflies.
+Here we will go through step-by-step how to interface with a swarm of crazyflies and make all the copters of the swarm hover and fly simultaneously in a square shape using the `Swarm()` class of the cflib.For this tutorial you will need a swarm (2 or more) of crazyflies with the latest firmware version installed and a global positioning system (Lighthouse,LPS or MoCap) that is able to provide data for the position estimation of the crazyflies. You can also use the Flowdeck but keep in mind that you should command relative movements of each Crazyflie and due to its nature it may lead to accumulative errors and unexpected behavior over time.
 
 # Prerequisites
 
@@ -16,7 +16,7 @@ We will assume that you already know this before you start with the tutorial:
 
 # Get the script started
 
-Since you should have installed cflib in the previous step by step tutorial, you are all ready to got now. Open up an new python script called `swarm_rectangle.py`. First you will start by adding the following import to the script:
+Since you should have installed cflib in the previous step by step tutorial, you are all ready to got now. Open up a new python script called `swarm_rectangle.py`. First you will start by adding the following import to the script:
 
 ```python
 import time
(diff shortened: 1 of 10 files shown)


docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,477 @@
+---
+title: "Step-by-Step: Swarm Interface"
+page_id: sbs_swarm_interface
+---
+
+Here we will go through step-by-step how to interface with a swarm of crazyflies and make all the copters of the swarm hover and fly simultaneously in a square shape using the `Swarm()` class of the cflib.For this tutorial you will need a swarm (2 or more) of crazyflies with the latest firmware version installed and a positioning system (Flowdeck,Lighthouse,LPS etc.) that is able to provide data for the position estimation of the crazyflies.
```
krichardsson, line 6: Should we be stricter and limit it to absolute positioning systems, that is remove the flow?
It will probably work with flow decks but it adds one more layer of complexity with relative positioning and accumulated errors.

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,477 @@
...
+We will assume that you already know this before you start with the tutorial:
+
+* Some basic experience with python
+* Followed the [crazyflie getting started guide](https://www.bitcraze.io/documentation/tutorials/getting-started-with-crazyflie-2-x/) 
+* Read the [high level commander](https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/api/cflib/crazyflie/high_level_commander/), [swarm](https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/api/cflib/crazyflie/swarm/) and [SyncCrazyflie](https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/api/cflib/crazyflie/syncCrazyflie/) documentation .
+
+
+# Get the script started
+
+Since you should have installed cflib in the previous step by step tutorial, you are all ready to got now. Open up an new python script called `swarm_rectangle.py`. First you will start by adding the following import to the script:
```
krichardsson, line 19: Typo: "an new" should be "a new"

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,477 @@
...
+    with Swarm(uris, factory=factory) as swarm:
+```
+
+This will import all the necessary modules and open the necessary links for communication with all the Crazyflies of the swarms.`Swarm` is a wrapper class which facilitates the execution of functions given by the user for each copter and can execute them in parallel or sequentially.Each Crazyflie is treated as a `SyncCrazyflie` instance and as the first argument in swarm wide actions.There is no need to worry about threads since they are handled internally.To reduce connection time,the factory is chosen to be instance of the `CachedCfFactory` class that will cache the Crazyflie objects in the `./cache` directory .
+
+The radio addresses of the copters are defined in the `uris` list and you can add more if you want to.
+
+# Step 1: Light Check
+
+In order to verify everything is setup and working properly a light check will be performed.During this check,all the copters will light up red for a short period of time and then return to normal.
```
krichardsson, line 48: Typo: There should be a space after "." and ",". This typo in seems to a here and there through out the text,

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,477 @@
...
+This is achieved by setting the parameter `led.bitmask` to 255 which results to all the LED's of each copter light up simultaneously.
+
+Add the helper functions `activateBitMask`,`deactivateBitMask` and the function`lightCheck` above `__main__`:
+```python
+def activateBitMask(scf):
```
krichardsson, line 53: These functions are not snake-case which is different from the names in the next section, should be the same I guess?
Also possibly rename to `activate_led_bit_mask` to make it clearer that it is related to LEDs and not any general bitmask.

krichardsson (changes requested):
Very nice tutorial! Great to explain how to use the Swarm class.
Some minor comments below.

docs/user-guides/sbs_swarm_interface.md
marios-stam: Clarified better in changes

krichardsson (approved):

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,481 @@
...
+
+# Step 1: Light Check
+
+In order to verify everything is setup and working properly a light check will be performed. During this check, all the copters will light up red for a short period of time and then return to normal.
+This is achieved by setting the parameter `led.bitmask` to 255 which results to all the LED's of each copter light up simultaneously.
+
+Add the helper functions `activate_led_bit_mask`,`deactivate_led_bit_mask` and the function`light_check` above `__main__`:
+```python
+def activate_led_bit_mask(scf):
+ scf.cf.param.set_value('led.bitmask', 255)
```
knmcguire, line 54: indenting problem here

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,481 @@
...
+with Swarm(uris, factory=factory) as swarm:
+        swarm.parallel_safe(lightCheck)
+        swarm.reset_estimators()
+```
+
+# Step 3: Taking off and Landing Sequentially
+Now we are going to execute the fist take off and landing using the high level commander. The high level commander is a class that handles all the high level commands like takeoff, landing, hover, go to position and others. You first need to enable it through the `commander.enHighLevel` parameter and then execute the take off and land commands through the below functions:
+```python
+def activate_high_level_commander(scf):
+    scf.cf.param.set_value('commander.enHighLevel', '1')
```
knmcguire, line 123: This is actually no longer required since of this PR: https://github.com/bitcraze/crazyflie-firmware/pull/903
 
The high level commander is always on, but just in a lower priority

We want to at one point remove the crtpcommanderhighlevel stop (https://github.com/bitcraze/crazyflie-firmware/issues/980), but that might break backwards compatibility

anyway, you don't need to mention it here.

docs/user-guides/sbs_swarm_interface.md
```diff
@@ -0,0 +1,481 @@
...
+# Step 2: Security Before Flying
+Before executing any take off and flight maneuvers, the copters need to make sure that they have a precise enough position estimation. Otherwise it will take off anyway and it is very likely to crash. This is done through `reset_estimators()` by resetting the internal position estimator of each copter and waiting until the variance of the position estimation drops below a certain threshold. 
+```python
+with Swarm(uris, factory=factory) as swarm:
+        swarm.parallel_safe(lightCheck)
+        swarm.reset_estimators()
+```
+
+# Step 3: Taking off and Landing Sequentially
+Now we are going to execute the fist take off and landing using the high level commander. The high level commander is a class that handles all the high level commands like takeoff, landing, hover, go to position and others. You first need to enable it through the `commander.enHighLevel` parameter and then execute the take off and land commands through the below functions:
```
knmcguire, line 120: perhaps an extra sentence why the high level commander is preferred (less communication,more autonomy on the crazyflie).  You can perhaps also refer somewhere to this doc that explans it: https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/sensor-to-control/commanders_setpoints/#high-level-commander but that can be also done on top.

knmcguire (changes requested):
Great work! It looks good in overal but I have some comments about the highlevel commander, regarding some outdated information. There is no need anymore to enable the high level commander separately.

docs/user-guides/sbs_swarm_interface.md
marios-stam: I had missed that completely. Maybe create a new issue for that in order to change it in the rest of the examples?

docs/user-guides/sbs_swarm_interface.md
marios-stam: Yes, you are right, thank you!

knmcguire (approved):

PR #353, opened 2022-08-10
remove opencv from cflib

knmcguire:
A proposal to check in the cflib if opencv is installed an do an except if not. 

The idea is to do approximately the same in the cfclient:

```
import pkg_resources
installed = {pkg.key for pkg in pkg_resources.working_set}
if {'opencv-python-headless'} - installed:
    OPENCV_INSTALLED = False
else:
    OPENCV_INSTALLED = True

```

and gray out the geometry opencv button if not installed.

I propose this to more or less not be as breaking case. In case person has an older cfclient but the latest cflib they get an exception (which won't close the client), and if they are up to date, the button will be gray outed.

This is towards closing this issue: https://github.com/bitcraze/crazyflie-clients-python/issues/611

### DIFF
--- a/cflib/localization/lighthouse_bs_geo.py
+++ b/cflib/localization/lighthouse_bs_geo.py
@@ -104,7 +104,7 @@ class LighthouseBsGeoEstimator:
             raise Exception('OpenCV is not installed. To use this function,' +
                             'do "pip3 install opencv-python-headless"' +
                             ' and restart the cfclient')
-        
+
         guess_yaw = self._find_initial_yaw_guess(bs_vectors)
         rvec_guess, tvec_guess = self._convert_yaw_to_open_cv(guess_yaw)
         rw_ocv, tw_ocv = self._estimate_pose_by_pnp(bs_vectors, rvec_guess, tvec_guess)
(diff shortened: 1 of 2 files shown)


knmcguire:
We should probably also remove this documentation as we are planning for it to be no longer experimental? https://www.bitcraze.io/documentation/repository/crazyflie-firmware/master/functional-areas/lighthouse/multi_base_stations/

krichardsson (commented):
Looking good! 

Maybe also add a function that can be used by the client to determine if the `LighthouseBsGeoEstimator` class can be used, something like `is_lighthouse_bs_geo_estimator_available()`

knmcguire:
Should that variable be part of the LighthouseBsGeoEstimator() class? The problem is is that the class is not initialized at startup of the dialog, but after the button is pressed and the angles are received: https://github.com/bitcraze/crazyflie-clients-python/blob/c10cd2499bfb8eea0677d69271397c84d393f85d/src/cfclient/ui/dialogs/lighthouse_bs_geometry_dialog.py#L183

Maybe if we have the initialization of the LighthouseBsGeoEstimator class in the init function of the dialog already so we can gray out the button immediately.

PR #356, opened 2022-08-18
Add outlier filter to LH geo estimation

krichardsson:
This pull request improves the lighthouse geometry estimation. It adds an outlier filter when estimating the initial guess. 
In some cases (mainly LH1) the initial guess could contain too many outliers, causing the refinement step to never converge.


knmcguire (approved):

knmcguire:
flight tested with the cfclient.  all looks good

PR #359, opened 2022-09-06
Fix Qualisys example

### LINKED ISSUE #166: Qualisys example script sensitive to errors when pushing full pose
The https://github.com/bitcraze/crazyflie-lib-python/blob/master/examples/qualisys/qualisys_hl_commander.py script gets the full pose from the QTM and pushes it into the Crazyflie using the send_extpose() function of the Extpos class in the python lib. For unknown reasons, the script seems to be sensitive to errors in the pose, resulting in a CF that flips or go in the wrong direction. It very much looks like the estimated position in the CF is wrong.

There are lots of possible error sources:
* Unexpected data from the Qualisys system
* A bug in the script when pushing the pose
* The std used when pushing the pose into the EKF might be off
* A bug in the firmware handling pose data
* Delays

This requires more investigations.

Using only the position works fine (through the send_extpos() method).

krichardsson:
The qualisys example has not been fully functional when using full pose. This PR fixes this.

Closes #166

### DIFF
--- a/examples/qualisys/qualisys_hl_commander.py
+++ b/examples/qualisys/qualisys_hl_commander.py
@@ -53,6 +53,10 @@ rigid_body_name = 'cf'
 # True: send position and orientation; False: send position only
 send_full_pose = False
 
+# When using full pose, the estimator can be sensitive to noise in the orientation data when yaw is close to +/- 90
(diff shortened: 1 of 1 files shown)


evoggy (approved):
Looks good!

PR #365, opened 2022-12-01
Catch errors when flashing decks

krichardsson:
Currently software errors that freezes the CF memory sub system are not handled in a good way. Typucally the flashing process hangs half way without any feedback.

This PR adds a timeout to the memory handing that calls the fail callback in case of a hang.


krichardsson:
We have decided to not use this solution and I'm closing this PR.
The problem should be solved in the firmware instead, see bitcraze/crazyflie-firmware#1160 as well as by handling disconnections better #366

PR #367, opened 2022-12-02
Handle disconnects when accessing memory

### LINKED ISSUE #366: Sync classes do not handle lost connection in a good way
There are a number of "Sync" classes that wraps the asynchronous nature of the python lib in a synchronous API. The behavior of these classes is not good when a Crazyflie is turned off for instance and they all need checked.

The `SyncCrazyflie` class is probably OK and it is the responsibility of the user to check that the connection still is alive when using the class, for instance by using the `is_link_open()` method, like this

``` python
with SyncCrazyflie(URI) as scf:
    while scf.is_link_open():
        do_something()
```

Other sync classes or methods, especially those that are blocking must return if the connection is lost to make it possible to exit. For instance the `write_sync()` method on the `DeckMemory` class. It may be used like this:  

``` python
with SyncCrazyflie(URI) as scf:
    deck = findDeckMemory()

    # write_sync() must return if connection is lost, otherwise the next line will hang forever
    deck.write_sync(0, large_blob_of_data)
```

Code like this can for instance be found in `cflib/bootloader/__init__.py` and currently this will hang if connection is lost while flashing a deck. 

It is probably a good idea to go through all sync implementations in the lib to make sure they are OK

krichardsson:
This PR adds fixes to detect and handle radio disconnects while accessing the mapped memory. 
If the Crazyflie is disconnected/turned off the failed callback will now be called to let the application know that the read/write failed. A future improvement could be to also add a reason for the failure.

Fixes #366

### DIFF
--- a/cflib/crazyflie/mem/__init__.py
+++ b/cflib/crazyflie/mem/__init__.py
@@ -256,6 +256,8 @@ class Memory():
         # Called when new memories have been added
         self.mem_added_cb = Caller()
 
+        self._clear_refresh_callbacks()
+
         # Called to signal completion of read or write
         self.mem_read_cb = Caller()
         self.mem_read_failed_cb = Caller()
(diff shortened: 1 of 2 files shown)


ataffanel (approved):
Good addition, every call with a callback should contain a fail callback if the main callback is not guaranteed to be called in case of error (similar to promises in JS ...). Good to add one to the memory subsystem.

PR #381, opened 2023-02-15
Removed use of commander.enHighLevel

### LINKED ISSUE #379: remove any highlevel commander start parameters in scripts
We still have examples that start the high level commander by a parameter, but the pushed based highlevel commander in the crazyflie -firmware no longer needs this.

krichardsson:
Remove use of the parameter `commander.enHighLevel`.

This parameter [is not used any more](https://github.com/bitcraze/crazyflie-firmware/pull/903)

Closes #379


knmcguire (approved):
Ah nice, it was even part of more scripts that I initially thought. Looks good

PR #382, opened 2023-02-15
Add full state setpoints

knmcguire:
So this adds the possibility to send full state set-point  from the CFlib to the Crazyflie. This is in the light of that the decoders on the crazyflie crtp receiving should always be reflected in the cflib.

This includes a

- send_full_state_setpoint() function
- a demo script that tests this
- renamed fp16.py in utils to encoding.py to make it more generic .

It's in draft for now. I'm not sure if the compressing really works, and I get a packet is too large error so got to look at this as well. Btw, I based the compressing function on the cpp library by @whoenig


knmcguire:
hmm if I use this full state setpoint for PID, it takes off, and it glides towards one direction, perhaps because the setpoint of roll pitch and yaw is 20 degrees in the beginning before showing the right setpoint at 0 degrees. It doesn't seem to be due to the compression. I haven't tested mellinger in flight yet but perhaps that doesn't happen? 

@whoenig is fullstatesetpoint suppose to work with PID as well or with mellinger alone?

whoenig:
I think it should work with PID. Internally all it sets is position control in the setpoint struct (with feed-forward terms for velocity/acceleration, which the PID controller might ignore).

knmcguire:
that's what I thought... not sure where this would come from then especially since with PID, the attitude setpoint will get overwritten anyway, or else I would have thought that the compressing still goes wrong somewhere. So more investigation is necessary

knmcguire:
so I confirm, for mellinger, this does not work either. So more investigation is necessary!

Next step: 
- unit tests for the compress decompress
- Test to see how the setpoints are changed on the crazyflie-firmware side.

tobbeanton (approved):

PR #383, opened 2023-02-16
Add notify setpoint

### LINKED ISSUE #301: Implement notifySetpointsStop packet
When switching from position setpoints to high level commander set points, it is necessary to implement the notifySetpointStop packet in the cflib api. 

see https://github.com/bitcraze/crazyflie-firmware/pull/903

knmcguire:
This adds the notify_setpoint_stop to the cflib 

I still need to flight test this, and we need to know if we bump the crtp version as well? 

This should be decoded by this part of the crazyflie-firmware
https://github.com/bitcraze/crazyflie-firmware/blob/674653717bede0f0d205a0e112585aa210109d3f/src/modules/src/crtp_commander.c#L94-L103

closes #301


krichardsson (approved):
Looking good!

PR #385, opened 2023-03-08
circling square demo with colors

hmllr:
First turn off LEDs, then take off as first thing in swarm cmd to avoid delays


knmcguire (changes requested):
see my comments above

krichardsson:
I might be a bit late in the game here, but I have a couple of improvements that I would like to suggest:
* Use pi (in math or numpy) instead of 3.14
* The copyright year in the header is a bit old :-)

PR #390, opened 2023-03-28
add wall following cflib demo

knmcguire:
We have had the [wall following app](https://github.com/bitcraze/crazyflie-firmware/tree/master/examples/demos/app_wall_following_demo) for a while in the crazyflie firmware. There used to be a python version made back when the SGBA work was done, but at one point all development was done using only the c version of the wall following algorithm.

I've now ported the c code back to python, so that we can have a cflib example of the wall following demo.


knmcguire:
Thanks for the tips! I've implemented a few already but I'll add issues to both the crazyflie-firmware and crazyflie-lib-python to implement these in the future. It would be good to keep the two versions synced up.

PR #410, opened 2023-07-26
Update sbs motion commander

krichardsson:
This PR corrects the documentation of the SBS motion commander.

The current text mentions that any positiong deck works, while the code only checks for the flow deck. This PR changes the text to only mention the flow deck


knmcguire (commented):
should we add a note somewhere, that another positioning deck can be used but that a different parameter needs to be checked?

Issue #411, opened 2023-08-10
Add other mocap examples minimal


knmcguire:
There are many users that uses Vicon and Optitrack but we only have examples for qualisys. It would be nice to also show an example for those others as well. We can either do this by using the SDKs directly or use the https://github.com/IMRCLab/libmotioncapture repo.

There is a issue open in the client repo already about this, but perhaps it can be argued to make this reliant on the cflib instead. https://github.com/bitcraze/crazyflie-clients-python/issues/577

This was the initial request by one of the users: https://github.com/bitcraze/crazyflie-lib-python/issues/151

whoenig:
+1 for adding it here. This could also lay the foundation to add broadcast support in cflib, which would resolve many painpoints in Crazyswarm2. And +1 for using libmotioncapture, rather than individual SDKs, because than you can also test it easily in-house with Qualisys and it should work without changed for vicon or optitrack users as well.

sesem738:
Hello,

I use an Optitrack for pose information. When I use vrpn, I get data, but when I use optitrack as the mocap_system_type, nothing happens. I don't know if it's something wrong on my side

knmcguire:
woops this one is still open! It shouldn't as an example have already been merged! See https://github.com/bitcraze/crazyflie-lib-python/pull/414

@sesem738 please start a discussion on https://discussions.bitcraze.io/ if you need support. Also mind that this example is based on the library in there: https://github.com/IMRCLab/libmotioncapture so it might also be a good step to first try out the minimal examples there.

PR #414, opened 2023-08-29
Add motioncapture lib example

krichardsson:
This PR adds an example script for using the motioncapture library to connect to a mocap system. Also moves the qualisys example to the new mocap directory


krichardsson:
Closes #411

PR #415, opened 2023-09-28
Update sbs_connect_log_param.md

PiggyGaGa:
simple_log function has wrong parameter logconf, which is not be used in the simple_log, the parameter name should be 'lg_stab' or change 'lg_stab' var as 'logconf' instead


knmcguire (changes requested):
Thanks for noticing, this is indeed a bug! 

Actually it should be the other way around. The function call should have logconf in there and the lg_stab in simple_log should be renamed logconf.

Could you also change it in the sbs code example as well? So to change lg_stab here to logconf:

 https://github.com/bitcraze/crazyflie-lib-python/blob/f35b0baf6ff243d618f7a8becf5a69b1ec6c1f56/examples/step-by-step/sbs_connect_log_param.py#L73

PR #417, opened 2023-10-10
Added usage message for radio-test

jath03:
It took me a while to figure out what I was doing wrong when trying to use the `radio-test.py` script.  I added a small usage comment to prevent other users from running into the same issue.


knmcguire:
Thanks! we will merge it.

knmcguire (approved):

PR #420, opened 2023-10-16
Correct Motion commander

### LINKED ISSUE #419: motion commanderf wo context manager fails on reconnect.
### Discussed in https://github.com/orgs/bitcraze/discussions/969

<div type='discussions-op-text'>

<sup>Originally posted by **dbaldwin** October 14, 2023</sup>
Hi,

I'm working my way through CRTP by using the motion_commander_demo.py example as a reference. I've modified the script to do a basic takeoff and land like this in this code snippet:

```
mc = MotionCommander(scf)
mc.take_off()
time.sleep(5)
mc.land()
```

It works well but for some reason if I run it again Crazyflie will ignore any packets. I have to restart my drone to get the script to work again. I've experienced this behavior with two different Crazyflies. Is there something I should be checking/resetting to get it to run subsequent times?

Thanks.</div>

krichardsson:
The motion commander uses low level setpoints and after landing we should hand over to the HL commander let it generate setpoins to avoid a timeout and locking of the Crazyflie. 

Closes #419

### DIFF
--- a/cflib/positioning/motion_commander.py
+++ b/cflib/positioning/motion_commander.py
@@ -117,6 +117,9 @@ class MotionCommander:
             self._thread = None
 
             self._cf.commander.send_stop_setpoint()
+            # Stop using low level setpoints and hand responsibility over to the high level commander to
+            # avoid time out when no setpoints are received any more
+            self._cf.commander.send_notify_setpoint_stop()
             self._is_flying = False
 
     def __enter__(self):


knmcguire:
Hmm I still find it strange that it is necessary... is the HL commander state the default?

But this is in light of previous changes you needed to do like this one. https://github.com/bitcraze/crazyflie-lib-python/pull/413/files

knmcguire (approved):

PR #439, opened 2024-01-26
Add support for upgrading nRF51 soft device and bootloader

evoggy:
This adds support for updating the nRF51 soft device and bootloader, which is needed so we can support the nRF51 softdevice update from s110 to s130. The softdevice and bootloader is distributed together as a binary, avoiding the risk of flashing incompatible softdeice/bootloader, which would require a debugger for restoring the nRF51. Below is a list of some of the changes made in the PR.

The following has been added to the manifest file:
* Bumped version to 2 (makes files incompatible with earlier version of the library)
* Added keys *provides* and *requires* to the file objects

The following was added for the bootloader communication:
* Version field update, used to report version of bootloader

Special handling has been added for the soft device flashing:
* Moving forward the s130+bootloader will be distributed in the release ZIPs, but in order to downgrade (i.e to an earlier release with s110 bootloader and nRF51 firmware) we need to have the s110+bootloader on the side since this has never been included in the ZIPs before. Therefore the binary for the s110+bootloader has been included in the library now.
* Special care is taken to not update the softdevice+bootloader over and over again if it doesn't need to be updated, the version (s110 or s130) is used together with the version of the bootloader (see above) to determine if softdevice+bootloader should be updated.
* Report to UI that we are flashing bootloader+softdevice


ataffanel:
Looks good, great job!

PR #442, opened 2024-03-10
Bluetooth Low Energy (BLE) Link Driver

UnexDev:
This PR implements the [CRTP over BLE](https://www.bitcraze.io/documentation/repository/crazyflie2-nrf-firmware/master/protocols/ble/). The implementation is not fully done yet, but I feel like opening this PR will allow for feedback and faster development.

The only thing that I don't know about is the fact that the only way to use the driver is to use the `asyncio` package along with it. This is, as far as I know, the only way to resolve a `coroutine`.

### DIFF
--- a/cflib/crtp/__init__.py
+++ b/cflib/crtp/__init__.py
@@ -33,6 +33,7 @@ from .serialdriver import SerialDriver
 from .tcpdriver import TcpDriver
 from .udpdriver import UdpDriver
 from .usbdriver import UsbDriver
+from .bledriver import BLEDriver
 
 __author__ = 'Bitcraze AB'
 __all__ = []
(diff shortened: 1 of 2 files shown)


knmcguire:
Hi! and thanks! Sorry for the late reply but we will discuss this at our next triage meeting

ataffanel:
We still needs some time to properly review it. It won't be time for the immediate next release though.

@UnexDev What is your current state on this PR? It seemed that it was still in progress. Do you want to commit anything else?

ToveRumar:
Hi @UnexDev !

I just want to check up on the status of this PR. 
Are you planning to keep working on it? 

If not we will need to close this PR for now, however it can be reopened if needed in the future.

ataffanel:
We will close this PR since the future will be with a Rust-backed crazyflie lib and we will then implement the BLE link (if any) in the Rust lib.

PR #444, opened 2024-03-30
cloader.read_flash is unreliable

dolfje:
When I tried to make a function that verified flash, it worked only 50% to have a correct page readout. Because the code only checks if the response is from the correct add, but not the correct part, it confuses the packets and add garbage in between.

change sponsored by stagebees.com


knmcguire:
Hi! thanks for this! I've marked it for our next triage meeting.

PR #454, opened 2024-05-07
pcap logging; Timestamps wrapping after 59 seconds.

ToveRumar:
The datetime object does not hold total seconds since Epoch time.The seconds variable is only seconds since last full minute. We need a deltatime since Epoch start to get total seconds since then

### DIFF
--- a/cflib/crtp/pcap.py
+++ b/cflib/crtp/pcap.py
@@ -1,6 +1,6 @@
 import os
 import struct
-from datetime import datetime
+import time
 from enum import IntEnum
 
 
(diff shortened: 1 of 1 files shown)


ToveRumar:
Fixes the issue from this discussion; 
https://github.com/orgs/bitcraze/discussions/1278

gemenerik:
Wireshark wiki confirms it should be seconds since Unix time. You could consider using time.time() since that returns Unix time in seconds.

ToveRumar:
> Wireshark wiki confirms it should be seconds since Unix time. You could consider using time.time() since that returns Unix time in seconds.

I liked that solution better. See update

gemenerik (approved):
Nice!

PR #459, opened 2024-06-27
Fix lighthouse geometry estimation example

gemenerik:
- Fixed the data collection reattempts. ensuring the loop runs again if necessary to successfully obtain recordings.
- Refactored code by consolidating data recording processes into dedicated functions for improved organization and readability.
- Implemented timeouts in the data recording functions to prevent indefinite blocking


ataffanel (approved):
Looks good!

PR #460, opened 2024-06-27
Add ip address instead of CRTP address to fpv example

ToveRumar:
This example needs to connect to the crazyflie over tcp.


gemenerik (approved):
Nice!

Issue #469, opened 2024-09-04
Implement quality stats


knmcguire:
Cpp link already has certain quality stats implemented that looks at amount of packets sent, how many acks have been returned and the delay of the messages being received (like here: https://github.com/bitcraze/crazyflie-link-cpp/pull/28). We should mirror this in cflib such that we can look at the communication in a more qualitative way. 

Currently broadcasting is not implemented in the cflib so we can't mirror that, but anything unicast we can ofcourse.

knmcguire:
Some extra information in the [connection.h of the statistics ](https://github.com/bitcraze/crazyflie-link-cpp/blame/master/include/crazyflieLinkCpp/Connection.h#L23). We should try to replicate these same stats

Currently I'm assuming that a connection statistic is per 1 crazyradio <-> 1 crazyflie pair

- sent_count: total amount of packets sent to a Crazyflie during the duration of the connection (so it never resets as long as it is connected)
- sent_ping_count: Total amount of empty packets sent _not sure what this is, related to broadcasting?_
- receive_count: total amount of packets received from the Crazyflie during the full connection
- enqueued_count: A sense of the queue of packets if packets are sent with safelink (which retries if packets aren't sent) _not exactly sure if that is actually the case_
- ack_count: amount of received acknowledgements acks from the Crazyflie

knmcguire:
In the cflib there are some link quality functionality already, which looks at the number of retries within a certain sliding window. So it returns a percentage of the retries done? 

https://github.com/bitcraze/crazyflie-lib-python/blob/be04130ac89acdbe7bd3b26684eb95db753ee564/cflib/crtp/radiodriver.py#L610-L618

Git blame and it has been implemented 8 years ago, but a typical thing that kind of dissappeard in memory.... but it has been used in the cfclient though! See number 3 in the main gui: https://www.bitcraze.io/documentation/repository/crazyflie-clients-python/master/userguides/userguide_client/#main-ui

I guess the question is, since we've got to leave this to prevent things from breaking, would we want to add some things on the side of this since cpp link has more values? Also broadcasting hasn't been implemented yet either so we can't add all (see #472 )

knmcguire:
Alright, so the thing here is, that cpp link looks at the full amount of packets send over the entire connection which is perhaps a bit much, but this looks at the retry percentage the last 100 packets which is way too little... But I kind of like the sliding window approach. 

[...]

PR #477, opened 2024-09-26
Introduce Link Statistics and Latency Measurement

gemenerik:
This PR introduces a framework for link statistics, enabling the tracking of various metrics, starting with latency measurement.

* LinkStatistics Class: A new class to manage multiple types of link statistics, facilitating the addition of more metrics in the future.
* Latency Measurement: A dedicated Latency class is introduced to send periodic ping requests to the Crazyflie. It calculates the round-trip latency and maintains the 95th percentile (p95) latency over recent measurements.

The implementation improves the lib's ability to monitor communication quality and lays the groundwork for further enhancements in link statistics.

### DIFF
--- a/cflib/crazyflie/__init__.py
+++ b/cflib/crazyflie/__init__.py
@@ -44,6 +44,7 @@ from .appchannel import Appchannel
 from .commander import Commander
 from .console import Console
 from .extpos import Extpos
+from .link_statistics import LinkStatistics
 from .localization import Localization
 from .log import Log
 from .mem import Memory
@@ -121,6 +122,7 @@ class Crazyflie():
         self.mem = Memory(self)
         self.platform = PlatformService(self)
         self.appchannel = Appchannel(self)
+        self.link_statistics = LinkStatistics(self)
 
         self.link_uri = ''
 
@@ -152,6 +154,11 @@ class Crazyflie():
             lambda uri: logger.info('Callback->Connection setup finished [%s]', uri))
         self.fully_connected.add_callback(
             lambda uri: logger.info('Callback->Connection completed [%s]', uri))
+
+        self.connected.add_callback(
+            lambda uri: self.link_statistics.start())
+        self.disconnected.add_callback(
+            lambda uri: self.link_statistics.stop())
 
     def _disconnected(self, link_uri):
         """ Callback when disconnected."""
@@ -283,6 +290,14 @@ class Crazyflie():
     def remove_port_callback(self, port, cb):
         """Remove the callback cb on port"""
         self.incoming.remove_port_callback(port, cb)
+    
+    def add_header_callback(self, cb, port, channel, port_mask=0xFF, channel_mask=0xFF):
+        """Add a callback to cb on port and channel"""
+        self.incoming.add_header_callback(cb, port, channel, port_mask, channel_mask)
+
+    def remove_header_callback(self, cb, port, channel, port_mask=0xFF, channel_mask=0xFF):
+        """Remove the callback cb on port and channel"""
+        self.incoming.remove_header_callback(cb, port, channel, port_mask, channel_mask)
 
     def _no_answer_do_retry(self, pk, pattern):
         """Resend packets that we have not gotten answers to"""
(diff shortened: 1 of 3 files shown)


gemenerik:
If you want to play around with it, I suggest printing in the `_ping_response` method

cflib/crazyflie/__init__.py
```diff
@@ -153,6 +155,11 @@ def __init__(self, link=None, ro_cache=None, rw_cache=None):
         self.fully_connected.add_callback(
             lambda uri: logger.info('Callback->Connection completed [%s]', uri))
 
+        self.connected.add_callback(
+            lambda uri: self.link_statistics.start())
+        self.disconnected.add_callback(
+            lambda uri: self.link_statistics.stop())
```
knmcguire, line 161: When using this library with the cfclient, upon reconnect with the cfclient still active, I get an error:

```
INFO:cflib.crazyflie:Callback->Connection setup finished [radio://0/40/2M/E7E7E7E705]
ERROR:cflib.crazyflie:Exception while doing callback on port [2]

Traceback (most recent call last):
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\__init__.py", line 441, in run
    cb.callback(pk)
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\param.py", line 548, in _new_packet_cb
    self._done_callback()
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\__init__.py", line 180, in _param_toc_updated_cb
    self.connected.call(self.link_uri)
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\utils\callbacks.py", line 54, in call
    cb(*args)
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\__init__.py", line 159, in <lambda>
    lambda uri: self.link_statistics.start())
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\link_statistics.py", line 68, in start
    self.latency.start()
  File "C:\Users\kimbe\Development\bitcraze\python\crazyflie-lib-python\cflib\crazyflie\link_statistics.py", line 107, in start
    self._ping_thread_instance.start()
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\threading.py", line 930, in start
    raise RuntimeError("threads can only be started once")
RuntimeError: threads can only be started once
```

But the .stop() does seem to close the thread, or at least it is intended to do that. So perhaps you can see if that actually happens.

knmcguire:
connected to #469

cflib/crazyflie/__init__.py
gemenerik: Good find! The threads were properly closed, but Python threading does not allow to start the same thread instance after stopping. Fixed with a6a92155d2b904f9a1e858368781dd6683cda7db.

knmcguire:
Anyway to attach a caller to this for this so that we can do a callback for each Crazyflie instance? Trying to find a good example to do that in a good way... the only one I could think of is the link_quality_callback but that's perhaps not the same level

gemenerik:
I would also like to discuss, before we merge, the ping interval, how many measurements we store, and if p95 is the best option

gemenerik:
> Anyway to attach a caller to this for this so that we can do a callback for each Crazyflie instance? Trying to find a good example to do that in a good way... the only one I could think of is the link_quality_callback but that's perhaps not the same level

Something like a `Caller` that you can attach a callback to for when a new latency value has arrived? I think `cflib/crazyflie/console.py` is a neat example of this. Or do you mean something else?

knmcguire:
Ah yes the console is a good option indeed. It's just that printing in a different thread gives issues sometimes, and I'd like to test it out with crazyswarm2 as well.

gemenerik:
> Ah yes the console is a good option indeed. It's just that printing in a different thread gives issues sometimes, and I'd like to test it out with crazyswarm2 as well.

Yeah I think adding a `Caller` could be neat, we can use it to update some UI feature in the client too. I can add it.

PR #479, opened 2024-09-26
Bandwidth congestion and rate from radio driver

knmcguire:
Just an example of how the congestion of the bandwidth can be calculated per null packet, along with the rate of the up and down link.

Not sure if this should actually be a call back of the crazyflie instance itself, since it is a statistic that is more linked to the crazyradio hardware as it looks at all the packets send. But the original link_quality was implemented like this though. 

This is connected to  #469

### DIFF
--- a/cflib/crazyflie/__init__.py
+++ b/cflib/crazyflie/__init__.py
@@ -208,6 +208,11 @@ class Crazyflie():
         """Called from link driver to report link quality"""
         self.link_quality_updated.call(percentage)
 
+    def _link_congestion_cb(self, congestion_up, congestion_down):
+        """Called from link driver to report link congestion"""
+        logger.info('Link congestion: up=%f down=%f', congestion_up,
+                     congestion_down)
+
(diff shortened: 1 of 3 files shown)


gemenerik:
If I understand correctly we need to count these packets at this low level, so the current implementation with a callback makes sense to me. Just a heads-up that counting null packets could be affected if we decide to "ease up" on them, as we discussed.

This is just a draft PR but some more implementation thoughts: If we’re using a callback, we might consider merging the low-level and existing quality callbacks into a single object with all the relevant information. I think the existing quality callback is passed to other links, but just not used, so we could do that here, too. Also, let’s consider moving all link statistics out of the run method into their own method (including the ack/retry quality stat).

knmcguire:
This is currently blocked until we are sure of the structure of #477

gemenerik:
Continued in #480

knmcguire:
Thanks! I'll remove the branch as well.

PR #480, opened 2024-10-03
Receive uplink RSSI

gemenerik:
### :rocket: Features
- Refactor link quality callback to contain signal health obj with additional signal health information
- Parse uplink RSSI in ack into signal health obj

### :memo: To-do
- CF link C++ uses a different calculation for link quality, investigate if we can unify this.
- Need to better integrate with #477, the signal health callback can probably be routed into LinkStatistics.
- Review naming for better clarity. Link statistics, link quality, and signal health can be easily confused. Also signal health is not a good name if we integrate things like bandwidth congestion.
- Can integrate bandwidth congestion #479 into the same signal health format.


gemenerik:
The C++ link driver's 'link quality' implementation is different from the (default) radio driver's. The the C++ link driver's implementation is based on the percentage of messages acked, while the radio driver's implementation is based on the number of retries.  The structure introduced in this PR is a work in progress, but given that they hook into the same callback, I may want to unify them.

gemenerik:
Latest commit combines #479 into this, updating PR to reflect

gemenerik:
Also, I verified, and the way this is set up currently, it does not break compatibility with `cfclient`

PR #482, opened 2024-10-04
Add callback for handling link errors in disconnected state

gemenerik:
Previously, there was no way to hook into link errors when the state was disconnected. 

For example, when somehow unable to connect to a CF (let's say because it was turned off); on the initial connection attempt, a callback would be triggered for a failed connection, allowing scripts to handle the error. However, subsequent attempts did not trigger any callback, forcing autonomous scripts to blindly retry with a timeout. This commit introduces a `disconnected_link_error` callback, enabling scripts to respond to link errors more effectively and retry connections programmatically.


tobbeanton (approved):
That is a really good addition. It all looks good for me.

PR #491, opened 2024-10-17
Modernize Python Packaging: Transition to pyproject.toml and PyPI Trusted Publishing

gemenerik:
* Switch from setup.py to pyproject.toml (PEP 518, PEP 621)
* Stop building (deprecated) universal wheels
* Use trusted publishing for PyPI

[Successful build, test PyPI publish](https://github.com/bitcraze/crazyflie-lib-python/actions/runs/11894137349)
[Successful build, PyPI publish](https://github.com/bitcraze/crazyflie-lib-python/actions/runs/11894292819)


gemenerik:
> Still requires testing. Also on non-Linux.

Tests are OK

PR #495, opened 2024-11-20
Dependency update, Py313 doc, and automatic versioning

gemenerik:
- Update dependencies (Numpy >1 breaks Vispy, so staying at 1)
- Add Python 3.13-specific install instructions (compile wheels, no Numpy 1 wheels available)
- Implement automatic versioning like in cfclient


ataffanel (approved):
LGTM

PR #504, opened 2024-12-16
Fix restart method in radiodriver.py.

h4writer:
Restart doesn't work anymore, since _RadioDriverThread expects a rate_limit. Fixing this by saving the rate_limit and reusing it here.


ToveRumar:
I just added a initialization of the rate_limit variable. Just for good measure. Otherwise it looks good. Ill approve and merge this PR. 

Thank you for your contribution!!

Issue #485, opened 2024-10-10
Clarify drones in swarm on same radio should be on same channel


gemenerik:
Drones in a swarm using the same radio must also be on the same channel for proper communication. While this is documented in some places, it’s easy to miss, especially if you're relying on examples rather than reading the documentation. You might modify URIs without realizing this and then seeing communication break down.

This is something that frequently pops up on GitHub discussions.

I suspect this confusion arises more often due to the nature of addressing. In our client, for instance, users are encouraged to differentiate drones by channel rather than radio address. However, for swarm communications to work correctly, this approach should be reversed.

A simple solution I see right now is to just add inline comments above all listed URIs in the examples.

PR #506, opened 2025-01-21
Added inline comments in the sbs swarm interface

ArisMorsinkPaloumpas:
Based on this issue [here](https://github.com/bitcraze/crazyflie-lib-python/issues/485), I added inline comments in the [step-by-step swarm interface](https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/user-guides/sbs_swarm_interface/), explaining that URIs in a swarm using the same radio must also be on the same channel.
I also added an extra sentence on the first script description explaining the same thing.


gemenerik (approved):
Nice!

PR #510, opened 2025-01-29
Adjust flashing sleep times to prevent AI-deck timeouts

gemenerik:
This PR adjusts the sleep times between flashing steps to prevent timeouts caused by the longer restart/boot time of the AI-deck introduced in https://github.com/bitcraze/crazyflie2-nrf-firmware/pull/102. When an AI-deck is attached, the sleep duration is increased accordingly.

Conversely, when no AI-deck is present, the sleep time is reduced by 1 second, aligning with the removal of the previous delay.


ataffanel (approved):
LGTM

Issue #323, opened 2022-03-07
Replace reset estimator in non-swarm examples


knmcguire:
The reset estimator has been converted for all swarm examples to a cflib function but not for the regular single drone examples... 

It would improve the cleanliness of all the examples, but might compromise the backward compatiblity that #288 was experiencing.

diegoch-it:
These are the examples to be modified:

knmcguire:
absolutely! Make sure to test it out though on at least one :) Thanks!

ArisMorgens:
Fixed in #515.

PR #515, opened 2025-02-11
Replaced reset estimator in single drone examples

ArisMorgens:
Converted the 'reset estimator' to a cflib function for the regular single drone examples so the scripts are getting smaller and look cleaner. This is based on this [issue](https://github.com/bitcraze/crazyflie-lib-python/issues/323). The new file is located in [crazyflie-lib-python/cflib/utils](https://github.com/bitcraze/crazyflie-lib-python/tree/master/cflib/utils).
# Modified Files:
## In autonomy:
1.  autonomousSequence.py
2.  autonomous_sequence_high_level.py
3. autonomous_sequence_high_level_compressed.py
4.  circling_square_demo.py (I removed the reset estimator because the swarm class is used.)
5.  full_state_setpoint_demo.py

## In lighthouse
1. lighthouse_openvr_grab.py
2. lighthouse_openvr_grab_color.py
3. lighthouse_openvr_multigrab.py

## In mocap
1. mocap_hl_commander.py
2. qualisys_hl_commander.py

## In positioning
1. flowsequenceSync.py
2. initial_position.py


cflib/utils/ResetEstimator.py
```diff
@@ -0,0 +1,56 @@
...
+            max_x = max(var_x_history)
+            min_y = min(var_y_history)
+            max_y = max(var_y_history)
+            min_z = min(var_z_history)
+            max_z = max(var_z_history)
+
+            if (max_x - min_x) < threshold and (
+                    max_y - min_y) < threshold and (
+                    max_z - min_z) < threshold:
+                print('reset estimator success!')
```
gemenerik, line 55: Maybe change this to be more in line with the `print('Waiting for estimator to find position...')`

cflib/utils/ResetEstimator.py
gemenerik, line 1: Although file naming in the `cflib` is not fully consistent, I would rename this file to `reset_estimator.py`. This also matches the naming convention for the other files in the utils folder.

examples/autonomy/circling_square_demo.py
```diff
@@ -72,55 +70,6 @@ def rotate_beizer_node(xl, yl, alpha):
     return x_rot, y_rot
 
 
-def wait_for_position_estimator(scf):
```
gemenerik, line 75: Do you think the estimator reset was not used on purpose or by accident?

gemenerik (changes requested):
Nice work!

examples/autonomy/circling_square_demo.py
ArisMorgens: This script is using the swarm class, so the reset_estimator can be done in a single line using swarm.reset_estimators(). There is no reason to use the single-drone reset_estimator function. It was probably used in this script before the swarm class implementation.

cflib/utils/ResetEstimator.py
ArisMorgens: Fixed in [this commit](https://github.com/bitcraze/crazyflie-lib-python/pull/515/commits/86df7156a64b3973d80d1a3fd2f0d38126898ed9).

PR #516, opened 2025-02-11
Use correct yaw convention in motion commander

### LINKED ISSUE #389: motion commander, start_linear_motion reverse yaw
If somebody uses the [start_linear_motion](https://www.bitcraze.io/documentation/repository/crazyflie-lib-python/master/api/cflib/positioning/motion_commander/#:~:text=def%20start_linear_motion(self%2C%20velocity_x_m%2C%20velocity_y_m%2C%20velocity_z_m%2C%20rate_yaw%3D0.0)) function from the motioncommander, and add a yaw_rate, you'll see that the resulting yaw rate on the crazyflie itself is reversed, compared to the coordinate system that we have given in our documentation: https://www.bitcraze.io/documentation/system/platform/cf2-coordinate-system/

This was originally introduced with this PR: https://github.com/bitcraze/crazyflie-firmware/pull/338 and implemented in all packets there forth. The problem is that the coordinate system is consistent with the state estimate but not consistent with the control input...

gemenerik:
This PR updates the motion commander to use the correct yaw convention for yaw rate, aligning with the [Crazyflie coordinate system](https://www.bitcraze.io/documentation/system/platform/cf2-coordinate-system/). Specifically, it ensures that yaw rate follows the right-hand rule, as defined in the Crazyflie documentation.

To achieve this, it requires reverting commit f6c0e737cfd061b72efcb125af7d545e06597807 in crazyflie-firmware. That commit inverted the received yaw rates upon arrival, which is no longer necessary. The corresponding change is implemented in [bitcraze/crazyflie-firmware#1456](https://github.com/bitcraze/crazyflie-firmware/pull/1456).

These two PRs must be merged together to avoid flipping the yaw rate (again). While we could consider maintaining compatibility, this was ultimately a bug—implemented incorrectly and inconsistently across different motion commander calls.

Fixes: #389

### DIFF
--- a/cflib/crazyflie/commander.py
+++ b/cflib/crazyflie/commander.py
@@ -26,6 +26,7 @@
 Used for sending control setpoints to the Crazyflie
 """
 import struct
+import warnings
 
 from cflib.crtp.crtpstack import CRTPPacket
 from cflib.crtp.crtpstack import CRTPPort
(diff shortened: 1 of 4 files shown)


ArisMorgens:
Looks good!

Issue #509, opened 2025-01-27
Add motion_commander example for two drones with different commands


ArisMorgens:
Based on the discussion [here](https://github.com/orgs/bitcraze/discussions/1749), it appears to be a bug where the motion commander does not interact properly with threading.
It would be a nice addition to have an example with two drones flying with different commands using the motion commander.

ArisMorgens:
Added example in #517.

PR #517, opened 2025-02-12
Added asynchronized swarm example with motion commander

ArisMorgens:
This is based on this [library issue](https://github.com/bitcraze/crazyflie-lib-python/issues/509).


gemenerik (approved):
Nice work on this! Given the number of support questions we've seen around swarming / threading and async execution, this should be really helpful for users.

PR #518, opened 2025-02-14
Added swarm-sharing-data example

### LINKED ISSUE #450: Add swarm example sharing data
Based on the discussion here: https://github.com/orgs/bitcraze/discussions/1277#discussioncomment-9186166

It would be nice to have an example of 2 crazyflies sharing data with eachother with the swarm class.

ArisMorgens:
It has been requested (e.g. in this [discussion](https://github.com/orgs/bitcraze/discussions/1277)) to add a swarm example where the position of each drone is available to the others in a swarm. This way a centralized collision-avoidance algorithm can be easily created.

In this script, a leader-follower scenario is implemented between 2 drones. One of them is the leader and performs its own independent trajectory, while the other one is the follower that is constantly commanded to keep a defined distance from the leader.

This fixes #450.

### DIFF
--- a/examples/swarm/leader-follower.py
+++ b/examples/swarm/leader-follower.py
@@ -1,0 +1,208 @@
+# -*- coding: utf-8 -*-
+#
+#     ||          ____  _ __
+#  +------+      / __ )(_) /_______________ _____  ___
+#  | 0xBC |     / __  / / __/ ___/ ___/ __ `/_  / / _ \
+#  +------+    / /_/ / / /_/ /__/ /  / /_/ / / /_/  __/
+#   ||  ||    /_____/_/\__/\___/_/   \__,_/ /___/\___/
+#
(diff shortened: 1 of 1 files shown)


examples/swarm/leader-follower.py
```diff
@@ -0,0 +1,208 @@
...
+
+if __name__ == '__main__':
+    cflib.crtp.init_drivers()
+
+    factory = CachedCfFactory(rw_cache='./cache')
+    with Swarm(uris, factory=factory) as swarm:
+
+        swarm.reset_estimators()
+
+        # swarm.parallel_safe(arm)
```
gemenerik, line 197: You can uncomment this so it works with the brushless out of the box. On platforms that are auto-armed sending `scf.cf.platform.send_arming_request(True)` again does not affect the behavior.

gemenerik (changes requested):

Issue #461, opened 2024-06-28
commander generic setpoints for manual flight


knmcguire:
Let's add some generic commander setpoints that is meant for manual flight, for both absolute roll/pitch and rate control.

The current way of how it is implemented in the Crazyflie is not ideal, since the response of rpyt setpoint can be changed by parameters which is a bit legacy. We don't want to remove that as that might break functionalities, but just an alternative way of sending rpyt setpoints that is more compatible with the other setpoints for velocity, hover position etc.

ArisMorgens:
This was fixed by adding a new generic commander packet in the firmware and a new setpoint command in the library.
The user can now control the thrust percentage instead of thrust and can switch between angle and rate using a bool variable.

PR #532, opened 2025-04-01
Added send_setpoint_manual command

ArisMorgens:
Currently, there are no generic commander setpoints, meant for manual flight.
Also, the current way of choosing between Angle and Rate mode for roll and pitch is a bit legacy. More specifically, you have to change the parameters in [crtp_commander_rpyt.py](https://github.com/bitcraze/crazyflie-firmware/blob/856d10ce4b9842cc7bdc7ed8bfdc7734991b1a49/src/modules/src/crtp_commander_rpyt.c#L71C1-L73C100) and re-flash the Crazyflie.

The `send_setpoint_manual` command allows the user to control the thrust percentage (from 0 to 100%) instead of thrust (from 1001 to 60000). Also, `rate` is a bool that switches the attitude mode input from Angle to Rate.

This is based on Library issue #461 and Firmware issue [#1388](https://github.com/bitcraze/crazyflie-firmware/issues/1388).

### DIFF
--- a/cflib/crazyflie/commander.py
+++ b/cflib/crazyflie/commander.py
@@ -47,6 +47,7 @@ TYPE_POSITION = 7
 TYPE_VELOCITY_WORLD = 8
 TYPE_ZDISTANCE = 9
 TYPE_HOVER = 10
+TYPE_MANUAL = 11
 
 TYPE_META_COMMAND_NOTIFY_SETPOINT_STOP = 0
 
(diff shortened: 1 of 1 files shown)


jb2197:
Hi, just to follow up on whether this function will be made available soon? Thank you!

ArisMorgens:
As an update, this PR was replaced by #539, which was successfully merged.

PR #533, opened 2025-04-02
Update the flake8 version.

ToveRumar:
it has been removed from the original repo.

We where using a very very old version so we where missing some features like global variable checking


ToveRumar:
The CI check is supposed to fail for this issue (global variables). 
Once the issue is fixed, i suggest we merge this

PR #538, opened 2025-04-14
Fix rotvec comparison to handle 180° rotation ambiguity

gemenerik:
Replaced direct rotvec comparison with relative rotation check to avoid sign ambiguity in 180° rotations. Fixes unit test failures on Windows on Python 3.10–3.12.


ataffanel (approved):
LGTM

PR #539, opened 2025-04-24
Added send_setpoint_manual command

ArisMorgens:
This is a follow up of [this firmware PR](https://github.com/bitcraze/crazyflie-firmware/pull/1470).
The `send_setpoint_manual` command allows the user to control the thrust percentage (from 0 to 100%) instead of thrust (from 1001 to 60000). Also, `rate` is a bool variable that switches the attitude mode input from Angle to Rate.

This is based on Library issue #461 and Firmware issue [#1388](https://github.com/bitcraze/crazyflie-firmware/issues/1388).


ataffanel (approved):
Great addition, looks good to me!

PR #543, opened 2025-06-05
Rework the lighthouse geometry estimator

krichardsson:
This PR contains changes that aims at reworking the lighthouse geometry estimator to make it easier to use and help the use to better understand the state of the system. 

This is work in progress.

This PR breaks APIs and it should be used together with bitcraze/crazyflie-clients-python#747 in the client.


enyanil (approved):
Thoroughly tested, looks really good!

PR #544, opened 2025-06-13
ParamUpdater thread should only run when there is a connection

ToveRumar:
In this PR I added functionality so that the paramupdater thread will start when a connection is initiated, not when the Param object is created. 
This makes the thread only live while there is a connection. 

Also make some other threads daemons so they close when main program closes.


ataffanel (approved):
LGTM, it is a nice step forward to clean the threading situation!

PR #547, opened 2025-06-26
Add support for trajectory high-level command with relative_yaw

gemenerik:
Adds support for new trajectory high-level command with relative_yaw. Chooses between legacy and new command format based on drone CRTP version, ensuring backward compatibility while enabling relative_yaw flag. 

Together with https://github.com/bitcraze/crazyflie-firmware/pull/1377


ToveRumar:
Is it possible to add relative yaw to the automous_sequence_hl_commander example so its easier to test and know how to use it?

ToveRumar (dismissed):
Looks good!

PR #569, opened 2025-09-18
Small example tweaks

gemenerik:
Standardize example filenames to snake_case
- Aligns with PEP 8 module naming conventions
- Improves consistency across examples directory

Added / lowered "landing" setpoints to prevent crash (detection)

Force Lighthouse deck into V2 mode for multi BS geometry estimation example


ataffanel (approved):
Looks great!

PR #574, opened 2025-10-17
Lighthouse Visualization Script

ArisMorgens:
This PR adds a simple script that reads the `.yaml` file of the lighthouse configuration and creates a 3D figure containing all the base stations. It can be useful for checking if your setup has good coverage of the flying space.

image


enyanil (approved):
Looks good to me, nice addition!

PR #604, opened 2026-08-17
Fix Lighthouse estimation documentation anchor

ataffanel:
## Summary

- Point the Initial estimation link at the generated `#2-initial-estimation-ippe` anchor.

## Why

The public website imports the `master` documentation and renders it with Kramdown's GFM parser. The previous fragment omitted the heading's numeric prefix and `IPPE` suffix, so it did not resolve. HTML-Proofer consequently blocked the `docker-int-web` release build.

## Validation

- Rendered the document with the `bitcraze/web-builder` GFM/Kramdown versions.
- Verified every same-page fragment resolves to a generated element ID.


ArisMorgens (approved):
Nice catch!

PR #606, opened 2026-08-20
Add default file to example persistent_params_from_file

stefanthorstenson:
Add default file to example `persistent_params_from_file.py`.

Tested:

```
$ python persistent_params_from_file.py
Persistent params: stored activeMarker.back!
```

### DIFF
--- a/examples/parameters/persistent_params_from_file.py
+++ b/examples/parameters/persistent_params_from_file.py
@@ -47,7 +47,7 @@ logging.basicConfig(level=logging.ERROR)
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('-f', '--file', type=str, default='params.yaml',
+    parser.add_argument('-f', '--file', type=str, default='persistent_params_from_file.yaml',
                         help='The yaml file containing the arguments. (default: params.yaml)')
     args = parser.parse_args()
 
--- a/examples/parameters/persistent_params_from_file.py
+++ b/examples/parameters/persistent_params_from_file.py
@@ -33,6 +33,7 @@ version: '1'
 """
 import argparse
 import logging
+import os
 
 import cflib.crtp
 from cflib.crazyflie import Crazyflie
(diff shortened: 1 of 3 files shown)


examples/parameters/persistent_params_from_file.yaml
```diff
@@ -0,0 +1,7 @@
+params:
```
ArisMorgens, line 2: Could you add a comment about what this file is about?
e.g. `# Used in persistent_params_from_file.py`

examples/parameters/persistent_params_from_file.py
```diff
@@ -40,16 +40,15 @@
...
 uri = uri_helper.uri_from_env(default='radio://0/80/2M/E7E7E7E7E7')
 
 # Only output errors from the logging framework
 logging.basicConfig(level=logging.ERROR)
 
-
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
-    parser.add_argument('-f', '--file', type=str, help='The yaml file containing the arguments. ')
+    parser.add_argument('-f', '--file', type=str, default='params.yaml',
```
ArisMorgens: A common way to run an example in cflib is to use uv from the root of the repo, since that's where the `pyproject.toml` is. (`uv run examples/parameters/persistent_params_from_file.py`). This won't work now because the file path is relative to the root. We could use something like `os.path.join(os.path.dirname(os.path.abspath(__file__)), 'params.yaml')` to fix that. What do you think?

ArisMorgens (commented):
Looks good! I left some comments that could reduce user confusion.

examples/parameters/persistent_params_from_file.py
stefanthorstenson: Good point! Will fix it.

examples/parameters/persistent_params_from_file.yaml
stefanthorstenson: Fixed. Added comment and changed name of the file to indicate relationship to python script.

examples/parameters/persistent_params_from_file.py
stefanthorstenson: Fixed.

ArisMorgens (approved):
Looks good!

PR #607, opened 2026-08-26
Us the same `link` instance in `send_packet()`

ArisMorgens:
When flashing a Crazyflie 2.0 via cfclient, most of the times the process got stuck in "identifying decks to update", with this error:
```
crazyflie-lib-python/cflib/crazyflie/__init__.py", line 390, in send_packet
    self.link.send_packet(pk)
    ^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'send_packet'
```

This was caused by `send_packet()` reading `self.link` twice: first in `if self.link is not None: `, and then in `elf.link.send_packet(pk)`. Between these reads, other threads (e.g. `close_link()`) can set the link to `None`.

This PR fixes that by reading self.link only once, so the commands in `send_packet()` will always agree.

### DIFF
--- a/cflib/crazyflie/__init__.py
+++ b/cflib/crazyflie/__init__.py
@@ -360,9 +360,10 @@ class Crazyflie():
             raise Exception('Data part of packet is too large')
 
         self._send_lock.acquire()
-        if self.link is not None:
+        link = self.link
+        if link is not None:
             if len(expected_reply) > 0 and not resend and \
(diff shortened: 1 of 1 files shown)


gemenerik (commented):
If another thread sets the link to None, isn't this supposed to fail?
I would say the problem is that we lose the connection while trying to identify the decks.

PR #610, opened 2026-09-09
Add FP16 param types

ratheron:
We've added FP16 param type to enable smaller floats to be transferred. We need this to transfer the floats for NN policies via cflib to the drone.


gemenerik (approved):
Thanks for the contribution! 

A TOC cached by an older cflib would need deleting. No unmodified firmware has ever had an FP16 param type, so nothing stale exists in practice.

